PR #2, opened 2013-08-07
enable the tf_prefix parameter

piyushk:
`tf_prefix` is being [deprecated](http://www.ros.org/wiki/hydro/Migration#tf2.2BAC8-Migration.Removal_of_support_for_tf_prefix) in ROS Hydro. I would still like to use the `tf_prefix` parameter http://www.ros.org/wiki/robot_state_publisher#Parameters to distinguish between different robots in multi-robot systems.

It seems that the robot_state_publisher code is not using either `tf::resolve` or reading in the `tf_prefix`, so I am a bit confused as to how this was was working pre-Hydro. I remember running some tests a few months ago that worked.

This patch reads the parameter explicitly and sets it using `tf::resolve`. Using this patch, I can run multiple robots in Gazebo without tf tree collisions.

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -56,7 +56,10 @@ JointStateListener::JointStateListener(const KDL::Tree& tree)
   // set publish frequency
   double publish_freq;
   n_tilde.param("publish_frequency", publish_freq, 50.0);
-  n_tilde.param("tf_prefix", tf_prefix_, std::string(""));
+  // get the tf_prefix parameter from the closest namespace
+  std::string tf_prefix_key;
+  n_tilde.searchParam("tf_prefix", tf_prefix_key);
+  n_tilde.param(tf_prefix_key, tf_prefix_, std::string(""));
   publish_interval_ = ros::Duration(1.0/max(publish_freq,1.0));
 
   // subscribe to joint state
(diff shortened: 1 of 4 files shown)


ugocupcic:
Is there any plan to integrate that in the default robot_state_publisher in hydro?

We're using the tf_prefix from robot_state_publisher a lot: for example publishing the targets to /targets/joint1 while having the robot positions on /joint1 etc... this way we can visualise easily in rviz.

advaitjain:
Would very much appreciate if this pull request can be merged.

isucan:
This looks good to me. @tfoote what do you think?

src/joint_state_listener.cpp
```diff
@@ -56,8 +56,8 @@
   // set publish frequency
   double publish_freq;
   n_tilde.param("publish_frequency", publish_freq, 50.0);
+  n_tilde.param("tf_prefix", tf_prefix_, std::string(""));
```
tfoote: To emulate previous behavior this should be searchParam

piyushk: @tfoote: nice catch. I am stuck in the middle of a few things right now. Will update PR later today and test it.

piyushk: fixed by @fsuarez6

piyushk:
bump

isucan:
bump

wjwwood:
@tfoote can you comment on this one?

tfoote:
This looks good to me.

isucan:
Should this be merged in hydro, or for indigo?

tfoote:
hydro

PR #6, opened 2013-11-19

jensenb:
This is dupe of https://github.com/ros/robot_state_publisher/pull/5 but the for groovy-devel branch


wjwwood:
Same thing as for hydro, I'm not sure if we should release this change backwards into groovy or not.

wjwwood:
I am going to close this and the master merge request as the person merging can manage the back porting if necessary, and I'm not even sure if the master branch is needed...

jensenb:
I only back ported the change to this branch because I wan unsure about what branch contains the active development, here on Github groovy-devel is listed as the main branch.

PR #7, opened 2013-11-19

jensenb:
This is dupe of https://github.com/ros/robot_state_publisher/pull/5 but for the master branch


wjwwood:
Closing, probably should drop the master branch, or we should merge to master and not hydro-devel or backport to hydro-devel. Either way closing until some decision is made for hydro.

PR #12, opened 2013-12-18
link to orocos_kdl explicitly

v4hn:
This fixes the following error:

Linking CXX executable
$HOME/ros/hydro/wet-ws/devel/lib/robot_state_publisher/state_publisher
/usr/bin/ld:
CMakeFiles/robot_state_publisher.dir/src/joint_state_listener.cpp.o: undefined
reference to symbol '_ZN3KDL7SegmentD1Ev'
$HOME/ros/hydro/wet/lib/liborocos-kdl.so.1.2: error adding symbols: DSO
missing from command line collect2: error: ld returned 1 exit status

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ project(robot_state_publisher)
 
 find_package(Eigen REQUIRED)
 
+find_package(orocos_kdl REQUIRED)
+
 find_package(catkin REQUIRED
   COMPONENTS roscpp rosconsole rostime tf tf_conversions kdl_parser
 )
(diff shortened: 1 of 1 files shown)


tfoote:
This is already in hydro-devel with a more complete patch c8b1d0cce5f27c30f019d09fa991da5f6a7b038e why are you using master?

v4hn:
counter question: Why is there a master branch at all
if it's not pointing to the latest development branch?

Thanks, I was a bit confused because I thought this was already
fixed but couldn't find it (in master..).

PR #13, opened 2014-02-26
Re-enabling rostests (a lot of them are broken because they've been disabled since fuerte)

jbohren:


isucan:
Does not seem to merge any more :( 
Do the tests pass?

jackieokay:
I'm revisiting this in my own branch targeting indigo (`fix_tests_indigo`), where I fixed the merge conflicts and updated the tf listeners to tf2. The tests pass individually if the configuration is correct (robot_description is set and robot_state_publisher gets started after the parameter is set to the correct value). But there's some orchestration problems with `catkin_make run_tests`: by default on my machine, the tests run in parallel, so `robot_description` gets overwritten and is wrong for some of the tests, causing them to fail. I also tried it with remapping `robot_description` and `joint_states` for `robot_state_publisher`, but the tests still fail in `catkin_make run_tests`.

https://github.com/ros/robot_state_publisher/compare/fix_tests_indigo

jackieokay:
new PR for re-enabling rostests in #41. I'll close this one.

Issue #19, opened 2014-09-27
Floating joints published as fixed joints


shaun-edwards:
The URDF specification allows for floating joint types (see [here](http://wiki.ros.org/urdf/XML/joint)).  However, the `robot_state_publisher` node does not recognize floating joint types.  Floating joints are instead interpreted as fixed joints.

This issue is related to ros/robot_model#86 (more information can be found there).

Issue #3, opened 2013-09-28
reenable rostests


dirk-thomas:
and after doing that please update the urls for test data from pr.willowgarage.com to download.ros.org.
- robot_state_publisher/CMakeLists.txt

jackieokay:
Fixed by #41

PR #20, opened 2014-10-13
Issue19 floating joints

shaun-edwards:
Addresses issue #19 and part of #3.  This PR depends upon ros/robot_model#87.

<EDIT>
PR is against `hydro-devel` (I need it in hydro), let me know if you want me to issue a PR against `indigo-devel` as well.

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -56,8 +56,11 @@ class JointStateListener{
 public:
   /** Constructor
    * \param tree The kinematic model of a robot, represented by a KDL Tree
+   * \param m Map of mimic joints
+   * \param ignore_joints List of joints to ignore when publishing transforms
    */
-  JointStateListener(const KDL::Tree& tree, const MimicMap& m);
+  JointStateListener(const KDL::Tree& tree, const MimicMap& m,
+                     const std::vector<std::string> ignore_joints);
 
   /// Destructor
   ~JointStateListener();
--- a/include/robot_state_publisher/robot_state_publisher.h
+++ b/include/robot_state_publisher/robot_state_publisher.h
@@ -63,8 +63,10 @@ class RobotStatePublisher
 public:
   /** Constructor
    * \param tree The kinematic model of a robot, represented by a KDL Tree 
+   * \param ignore_joints List of joints to ignore when publishing transforms
    */
-  RobotStatePublisher(const KDL::Tree& tree);
+  RobotStatePublisher(const KDL::Tree& tree,
+                      const std::vector<std::string>ignore_joints);
 
   /// Destructor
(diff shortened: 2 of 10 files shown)


shaun-edwards:
ping?

include/robot_state_publisher/joint_state_listener.h
```diff
@@ -56,8 +56,11 @@ class JointStateListener{
 public:
   /** Constructor
    * \param tree The kinematic model of a robot, represented by a KDL Tree
+   * \param m Map of mimic joints
+   * \param ignore_joints List of joints to ignore when publishing transforms
    */
-  JointStateListener(const KDL::Tree& tree, const MimicMap& m);
+  JointStateListener(const KDL::Tree& tree, const MimicMap& m,
+                     const std::vector<std::string> ignore_joints);
```
isucan, line 63: could this be a const &?

include/robot_state_publisher/robot_state_publisher.h
```diff
@@ -77,7 +79,8 @@ class RobotStatePublisher
   void publishFixedTransforms(const std::string& tf_prefix);
 
 private:
-  void addChildren(const KDL::SegmentMap::const_iterator segment);
+  void addChildren(const KDL::SegmentMap::const_iterator segment,
+                   const std::vector<std::string>ignore_joints);
```
isucan, line 83: const &?

isucan:
This is an ABI change too (for Hydro). However, this code is not widely liked against, as far as I know. I am fine with merging. @tfoote is this ok with you as well?

shaun-edwards:
@isucan, thanks for the review.  I can make the desired changes.  Is there a way to make my changes ABI compatible?  If I include a wrapper method with the same signature as before?

If not, I'm OK just including this in Indigo.  This change is to support our calibration package, which is only a source release in Hydro.

isucan:
That's a good point. Maybe we can keep the old constructor, and that is the same as an empty ignore_joints. Ping @tfoote @dirk-thomas

tfoote:
Maintaining the old constructor will keep API compatibility, and should be done. ABI compatibility is hard to confirm without a checker. We're recompiling everything anyway at the moment so ABI compatibility is not strictly required. Also this is mostly used as an executable not a library.

jackieokay:
ok, in order to close ros/robot_model#86 I'd like to close this issue and do something with this set of changes. Can I target it to Kinetic at least? I'll get an ABI checker to confirm everything is ok

shaun-edwards:
My fault for not responding to @tfoote's request.  This issue may have been fixed in this #32 (already merged).  If so, I'm pretty sure this PR and it's changes are no longer needed.

jackieokay:
no problem. I can't reproduce the error in Indigo, so I'll close this unless someone suggests otherwise.

PR #26, opened 2015-03-30
get rid of argv[0] debug output on startup

xqms:
`robot_state_publisher` prints `argv[0]` on startup, which is probably a leftover debug statement. Simply remove it. We have enough log clutter as it is ;-)


wjwwood:
+1

PR #28, opened 2015-06-24
Port to tf2 and enable using static broadcaster

paulbovbel:
Depends on https://github.com/ros/geometry_experimental/pull/114

Added a "use_tf_static" parameter, which enables broadcasting static transforms over /tf_static using a latched publisher.

### DIFF
--- a/include/robot_state_publisher/robot_state_publisher.h
+++ b/include/robot_state_publisher/robot_state_publisher.h
@@ -40,8 +40,9 @@
 #include <ros/ros.h>
 #include <boost/scoped_ptr.hpp>
 #include <tf/tf.h>
-#include <tf/transform_broadcaster.h>
 #include <urdf/model.h>
+#include <tf2_ros/static_transform_broadcaster.h>
+#include <tf2_ros/transform_broadcaster.h>
 #include <kdl/frames.hpp>
 #include <kdl/segment.hpp>
 #include <kdl/tree.hpp>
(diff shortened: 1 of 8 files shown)


adolfo-rt:
This is a great enhancement!. Do you have some numbers of `rostopic bw /tf` with/without the feature?. If you do, please accompany the data with the total number of joints and the number of fixed joints.

paulbovbel:
@adolfo-rt, the reduction is linear with the number of fixed joints. From what I remember, each fixed joint eats up approx 5-10 KB/s, at the default publish rate (50 Hz).

paulbovbel:
Upstream changes to geometry_experimental have been merged and released, ping for review.

adolfo-rt:
> @adolfo-rt, the reduction is linear with the number of fixed joints. From what I remember, each fixed joint eats up approx 5-10 KB/s, at the default publish rate (50 Hz).

I just checked, and our REEM robot has +100 joints, of which almost 40% are fixed. So the bandwidth improvements should be significant in such a case. Good work!.

paulbovbel:
Glad to share the bandwidth reduction - this was a low hanging fruit, and I'm surprised it never happened during the tf2 transition.

paulbovbel:
Ping @isucan, can you please review?

isucan:
This looks good to me. The only concern I have is whether this should be merged to jade, not indigo. The improvements are great, but I worry about the ABI change, given this is such a low level package. What do you think @wjwwood ?

paulbovbel:
Either way, in jade I would advocate for making 'use_static' the default behaviour.

wjwwood:
As long as the default behavior is not changing in Indigo, that's probably fine. I'm a bit wary to change the default behavior in Jade, but since it should be a transparent change I'd probably go ahead with it.

paulbovbel:
@isucan, is this okay to merge as is? I can write a separate pull to jade-devel with this behaviour as default.

isucan:
Yes, I think we can merge this, but there seem to be some conflicts (related to another change that got pushed). We should also keep the jade branch in sync, and add the change to the default you mentioned (in jade).

jackieokay:
I rebased on `indigo-devel` and fixed merge conflicts in `origin/tf2-static`:

https://github.com/ros/robot_state_publisher/tree/tf2-static

@paulbovbel if that looks correct to you, you can either rebase your branch on the upstream branch, or we can just open a new PR with the rebased branch if you don't have time.

I will also open a separate PR that targets jade.

paulbovbel:
thanks @jacquelinekay 

@isucan, this PR is ready to go

include/robot_state_publisher/robot_state_publisher.h
```diff
@@ -63,27 +64,32 @@ class RobotStatePublisher
...
   void publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time, const std::string& tf_prefix);
-  void publishFixedTransforms(const std::string& tf_prefix);
+  void publishFixedTransforms(const std::string& tf_prefix, bool use_tf_static);
 
 private:
   void addChildren(const KDL::SegmentMap::const_iterator segment);
 
 
   std::map<std::string, SegmentPair> segments_, segments_fixed_;
+<<<<<<< HEAD
```
rethink-imcmahon, line 86: Yup. Sure enough, the merge conflict is here. Good catch, @remod. Did this branch need to be rebased before merging?

rethink-imcmahon, line 86: This is fixed by #36

PR #30, opened 2015-07-04
add option for setting namespace as tf_prefix automatically

cyborg-x1:
I would propose this to be able to set the tf_prefix automatically instead of supplying it in the launchfile manually.
According to http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix is deprecated, but currently I do not see another approach for having multiple robots.

Topics seem to do this automatically but only robot state publisher is somehow missing the feature to adapt to the namespace inside a group automatically.

I am trying to get my launchfile a bit more clean that I can put only one launchfile and just call it in multiple groups to achieve the same result. But robot state publisher is preventing this by not adapting the namespace automatically. Or maybe I am missing an option in string supplied to tf_prefix, if so, just tell me and forget about this.

Actually a namespace environment variable in launchfile could also do the same trick, but  there seems to be none available, if there is a way to get the current namespace into a parameter, please tell me, then this would also get obsolete.

I got only one way around this. putting the namespace into an argument. By default it would be ns="/" so it works with both groups and robot state. The disadvantage is that you have to have a group inside the launchfile itself, that wouldn't be the case with this little change.

I also found out that removing  a leading "/" is not necessary, it is automatically done by the following functions.

Now tell me what you think, or how it is supposed to be done.

Regards,

Christian

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -56,10 +56,23 @@ JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m)
   // set publish frequency
   double publish_freq;
   n_tilde.param("publish_frequency", publish_freq, 50.0);
-  // get the tf_prefix parameter from the closest namespace
-  std::string tf_prefix_key;
-  n_tilde.searchParam("tf_prefix", tf_prefix_key);
-  n_tilde.param(tf_prefix_key, tf_prefix_, std::string(""));
+
+  bool getns;
+  n_tilde.param<bool>("ns_tf_prefix", getns, false);
+
+  /*get current namespace as prefix for tf*/
+  if(getns)
+  {
+	  tf_prefix_=n.getNamespace();
+  }/*tf_prefix behavior*/
+  else
+  {
+	  // get the tf_prefix parameter from the closest namespace
+	  std::string tf_prefix_key;
+	  n_tilde.searchParam("tf_prefix", tf_prefix_key);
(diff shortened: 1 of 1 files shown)


Rayman:
This is a good question. In ros-simulation/gazebo_ros_pkgs#333 they use xacro to solve this (dynamically generating the launch file).

Maybe you could add an option to enable this behavior?

clalancette:
It's been quite a while since someone looked at this; I'm sorry about that.  Is this still a problem you are struggling with?  Have you found workarounds in the meantime?

sloretz:
@cyborg-x1 Because the tf_prefix has been deprecated I don't think it should be added to robot_state_publisher.

I think you can make do with namespacing. A way of launching multiple robots with a parameterized launch file is to remap the topics "/tf" and "/tf_static" from fully qualified names to relative ones: "tf" and "tf_static". Then the tf topics can then be pushed into a namespace "/robot1/tf" and "/robot1/tf_static". 

This isn't perfect. If you want to see all transforms for all robots you'll need to make a bridge between each of the robot tf topics and "/tf" + "/tf_static" topics. A bridge might see `map->odom` on topic "/robot1/tf" and publish `robot1_map->robot1_odom` on "/tf". If you launch Rviz you'll have to either remap the global tf topics to a robot specific one, or republish all messages that use `std_msgs/Header` with an updated frame_id.

There isn't a perfect option for multi-robot transforms, but I think it would be better discussed on ros/geometry2#32. I'll close this for now, but feel free to reopen.

PR #32, opened 2015-09-01
Check URDF to distinguish fixed joints from floating joints. Floating…

### LINKED ISSUE #19: Floating joints published as fixed joints

toliver:
… joints are ignored by the publisher (no segment is added for them).

Fixes #19 

Maybe [this](https://github.com/ros/robot_model/blob/indigo-devel/kdl_parser/src/kdl_parser.cpp#L88) warning should be modified to make it clear that it happens at the KDL level only, but the robot_state_publisher is still able to distinguish.

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -57,7 +57,7 @@ public:
   /** Constructor
    * \param tree The kinematic model of a robot, represented by a KDL Tree
    */
-  JointStateListener(const KDL::Tree& tree, const MimicMap& m);
+  JointStateListener(const KDL::Tree& tree, const MimicMap& m, const urdf::Model& model = urdf::Model());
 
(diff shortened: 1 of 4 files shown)


ugocupcic:
Hi,

Anything we can do to help speeding up that merge?

Cheers,

isucan:
This looks good, but can you make the request against the jade branch?

isucan:
On further consideration and discussion with @wjwwood, this has a low chance of causing problems, so merged in indigo. Thank you for your fix!

ugocupcic:
OK great thanks for the merge!

PR #34, opened 2015-10-14
[jade] Port to tf2 and enable using static broadcaster

jackieokay:
Branch of #28 targeting Jade.


paulbovbel:
LGTM, pending #35

PR #39, opened 2015-12-09
Fix API break in publishFixedTransforms

### LINKED ISSUE #38: API incompatibility in RobotStatePublisher::publishFixedTransforms
Commit 8c27d2b09f43726340974e952ae646ba9601ca08 broke backwards compatibility of the C++ interface by adding a `bool use_tf_static` parameter to the `RobotStatePublisher::publishFixedTransforms` member function.

Adding a default value to either true or false will fix the incompatibility. If filling in a default is undesirable, we could add a deprecated overload that fills in the boolean value.

Either way, breaking the C++ API within a the same ROS release is a bit messy.

scpeters:
A bool argument was added to
RobotStatePublisher::publishFixedTransforms
which broke API.
I've added a default value of false, to match
the default specified in the JointStateListener
constructor.

Fixes #38.

### DIFF
--- a/include/robot_state_publisher/robot_state_publisher.h
+++ b/include/robot_state_publisher/robot_state_publisher.h
@@ -76,7 +76,7 @@ public:
    * \param time The time at which the joint positions were recorded
    */
   void publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time, const std::string& tf_prefix);
-  void publishFixedTransforms(const std::string& tf_prefix, bool use_tf_static);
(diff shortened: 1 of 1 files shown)


scpeters:
This is targeted to `jade-devel`, but it can be back-ported to `indigo-devel`

jackieokay:
+1

jackieokay:
I'll let you back-port it (you have push access to this repo now, right?)

scpeters:
Yes, I'll do it

scpeters:
indigo-devel: 232fb691da5540c3b8a7f73b88a5467561b2ee00

Issue #23, opened 2015-01-23
no public default constructor.


Karsten1987:
is there any reason for not providing any public default constructor? This prevents any declaration of a robot state publisher as a member variable and initialize it on runtime.

jackieokay:
No idea. I'll implement this for Kinetic.

jackieokay:
:turtle:

jackieokay:
hey @Karsten1987, if you're still interested in this feature, I made a potential implementation for the default constructor of `RobotStatePublisher` in #43. I can also add one for `JointStateListener`.

cheers!

clalancette:
@Karsten1987 Are you still interested in this issue?  If so, we can probably pretty easily push it over the finish line.  If not, I'll close it out.  Just let me know.

Karsten1987:
i like that once a year, we push that issue forward :D 

I am currently don't have any code which uses it, but I still think the proposal is valid.

clalancette:
Haha, nice.  Sure, I'll leave it open, and hopefully get to it soon :).

clalancette:
I'm going to close this out.  This is unlikely to ever be done for ROS 1, and the whole thing is rewritten in ROS 2.

PR #43, opened 2016-01-24
Implement public default constructor

jackieokay:
Potential implementation of the enhancement discussed in #23. The default constructor of `RobotStatePublisher` creates an instance of the class with an empty KDL tree and an empty URDF model. The `init` function is added for delayed initialization of the KDL tree.


jackieokay:
Proper implementation of this delayed `init` feature will require removing the `const` qualifier from the `model_` member of RobotStatePublisher.

clalancette:
Since this is pretty old, and I now have a newer PR over at #87, I'm going to close this out in favor of that one.

Issue #25, opened 2015-02-24
Missing support for subclassing


peci1:
I'd like to customize the behavior of both joint_state_listener and robot_state_publisher, but the interface doesn't allow for that. In joint_state_listener, all the methods are private. And both in joint_state_listener and robot_state_publisher, no method is virtual to support subclassing. Furthermore, there's no way to "plug in" a different class than RobotStatePublisher to the joint_state_listener.

I'd like to ask you to make this package more friendly for subclassing.

If you'd like to know my motivation for subclassing - in our project, we're forced to fake some fixed joints by prismatic joints with zero limits. And I'd like the robot_state_publisher to recognize these fake fixed joints as fixed and to treat them so. And I'd also like to use the /tf_static topic for publishing the transforms for fixed joints.

jackieokay:
hey @peci1,

it's a bit late, but I can look into adding better support for subclassing in Kinetic.

peci1:
@jacquelinekay  +1

peci1:
Thanks ;) Is there a possibility to get it also merged into indigo-devel?

jackieokay:
I believe it technically isn't an API change and could go into Indigo, but it will definitely break ABI. I will think about it.

PR #45, opened 2016-02-18
Support for subclassing

jackieokay:
Connects to #25 

@peci1 I think this is ready to get into Kinetic. What do you think?


peci1:
Looks nice :) I haven't tested that, but if you wrote the tests, I think we can believe it. Thanks ;)

PR #47, opened 2016-05-11
Remove use_tf_static and backwards compat. for non-static fixed transforms

jackieokay:
feedback from @paulbovbel, @ros/ros_team, @ros/robot_model all welcome.

I've decided to remove the backwards compatibility for publishing fixed transforms that do not use tf static in Kinetic, since semantically I don't think there should be a difference between different kinds of fixed transforms and static transforms are more efficient.

Tully brought up that the only issue is that static transforms are usually not captured by rosbag recording because they are latched topics. He also mentioned that one or two introspection tools might not be aware of static transforms but wasn't specific about which ones.

### DIFF
--- a/include/robot_state_publisher/robot_state_publisher.h
+++ b/include/robot_state_publisher/robot_state_publisher.h
@@ -76,7 +76,7 @@ public:
    * \param time The time at which the joint positions were recorded
    */
   virtual void publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time, const std::string& tf_prefix);
-  virtual void publishFixedTransforms(const std::string& tf_prefix, bool use_tf_static = false);
+  virtual void publishFixedTransforms(const std::string& tf_prefix);
 
 protected:
   virtual void addChildren(const KDL::SegmentMap::const_iterator segment);
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -56,8 +56,6 @@ JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m,
   // set publish frequency
   double publish_freq;
   n_tilde.param("publish_frequency", publish_freq, 50.0);
-  // set whether to use the /tf_static latched static transform broadcaster
-  n_tilde.param("use_tf_static", use_tf_static_, false);
   // get the tf_prefix parameter from the closest namespace
   std::string tf_prefix_key;
   n_tilde.searchParam("tf_prefix", tf_prefix_key);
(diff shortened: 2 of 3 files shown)


jackieokay:
@ros-pull-request-builder Test this please

wjwwood:
Bag files do capture latched topics afaik.

dirk-thomas:
@jacquelinekay You might want to create a pull request (https://github.com/ros/rosdistro/blob/4564d744411db34af4a899b6e467dfa986246536/kinetic/distribution.yaml#L1894-L1897) otherwise nothing is going to happen :wink:

jackieokay:
yeah, I just noticed that

jackieokay:
ros/rosdistro#11498

paulbovbel:
It'd be an interesting idea to make this the default for kinetic, but leaving backwards compatibility as on option can't be a bad thing.

rosbags not capturing latched topics is definitely one big reason. Another is that rospy tf(1) does not use tf2 under the hood, so anything built on that will stop working.

wjwwood:
I just checked, rosbag definitely records latched topics and plays them back as latched. There's no issue with rosbag and using the latched topic.

What you might need is a way to migrate old bag files recorded in the other style though. It's not clear to me that this is necessary though.

tfoote:
@wjwwood It does do latching correctly for a single source, but it does not reproduce the logic if there's two publishers: http://answers.ros.org/question/207551/how-to-save-static-transforms-in-bag-files/?comment=207942#comment-207942

There's also a useability challenge when recording bags that are split, the split bags do not rerecord the static data which means that subsequent bags cannot be used independently. 

Old bags with static transforms published as future dated dynamic ones will work as before.

wjwwood:
I see the problem you're describing, but when talking about the issue we should describe it as such rather than saying "rosbag doesn't capture latched topics". It also captures the latched topics from multiple sources it just doesn't emulate the fact that there were multiple sources on play back. 

Are there ROS comm or tf issues to track these problems? It would be good to link to them here when talking about the problem.

jackieokay:
If the issues with rosbag do not get resolved before Kinetic (this week) I won't merge this and opt to cherry-pick #50 instead. It sounds like the problems with latched topics in rosbag have been hanging around for a while and should be addressed.

wjwwood:
In summary, there are two issues:
- rosbag's which are split during recording only contain latched messages in the first bag
  - Work around is to playback all bags and jump to a time index you want, but that's a bad UX.
  - Existing issue: https://github.com/ros/ros_comm/issues/706
- rosbag cannot generically emulate multiple publishers on a latched topic
  - Workaround for `/tf_static` is to remap it out of the `rosbag play` to `/tf_static_remap` or similar, have a node that subscribes to the remapped topic, collect messages published and republish them as a block of messages on `/tf_static`.
  - Existing issue: https://github.com/ros/ros_comm/issues/146

jackieokay:
Alright, I'm going to close this for now since it sounds like the rosbag issues won't get resolved in the near future.

PR #53, opened 2016-06-01
added repitition to robot_description reading

### LINKED ISSUE #52: robot_state_publisher fails if the robot_description is not immediately on parameter server
I have a system that generates the urdf based on a hardware scan. Ideally robot_state_publisher should retry a few times (or after a given delay) before failing.

VahidAminZ:
fixes #52

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -62,6 +62,12 @@ public:
   /// Destructor
   ~JointStateListener();
 
+  // Default value for number of attempts to read robot_description from parameter server
+  static const int default_description_read_repetitions_;
+
+  // Default value for delay between the attempts to read robot_description
+  static const double default_description_read_delay_;
+
 protected:
   virtual void callbackJointState(const JointStateConstPtr& state);
   virtual void callbackFixedJoint(const ros::TimerEvent& e);
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -46,6 +46,8 @@ using namespace std;
 using namespace ros;
 using namespace KDL;
 using namespace robot_state_publisher;
+const int JointStateListener::default_description_read_repetitions_ = 0;
+const double JointStateListener::default_description_read_delay_ = 0.5;
 
 JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m, const urdf::Model& model)
   : state_publisher_(tree, model), mimic_(m)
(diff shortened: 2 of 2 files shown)


include/robot_state_publisher/joint_state_listener.h
```diff
@@ -62,6 +62,12 @@ class JointStateListener {
   /// Destructor
   ~JointStateListener();
 
+  // Number of tries for reading robot_description from parameter server
+  static const size_t description_read_repitions_;
```
DLu: Change to description_read_repititions_?

VahidAminZ: fixed.

include/robot_state_publisher/joint_state_listener.h
```diff
@@ -62,6 +62,12 @@ class JointStateListener {
   /// Destructor
   ~JointStateListener();
 
+  // Number of tries for reading robot_description from parameter server
+  static const size_t description_read_repititions_;
```
jackieokay: nitpick: please spell "repetitions" correctly

jackieokay:
I think that the best implementation of this feature would be to add description_read_repetitions and description_read_delay as ROS parameters with default values if the parameters are not set on the server.

VahidAminZ:
@jacquelinekay Makes sense. I changed it accordingly. I am not sure about default values though.

jackieokay:
I would consider making the default value for `default_descriptions_read_repetitions` 0. That way users can opt into the new behavior.

@ros/ros_team and @ros/robot_model, do you have any feedback on this new workaround for nondeterministic startup behavior?

VahidAminZ:
@jacquelinekay changed the default repetitions to 0.

toliver:
Hi @jacquelinekay just checking if this PR can be merged or there is some fundamental disagreement on the content.

If it's going to be merged we can resolve the conflicts.

toliver:
This has been rebased and the checks are passing. Do you think it can be merged?

include/robot_state_publisher/joint_state_listener.h
```diff
@@ -62,6 +62,12 @@ class JointStateListener {
   /// Destructor
   ~JointStateListener();
 
+  // Default value for number of attempts to read robot_description from parameter server
+  static const int default_description_read_repetitions_;
```
sloretz, line 66: I think adding these two constants to the JointStateListener class breaks ABI, which would block merging into kinetic-devel. Since these are only used in main(), how about they be local variables there?

sloretz:
Hi @toliver, thanks for the contribution. If I understand correctly this enables robot_state_publisher to wait for the parameter "robot_description" to be set for a fixed amount of time.

Would you mind adding a test to make sure it does not wait by default, and a test to make sure it can wait when the two parameters are set?

src/joint_state_listener.cpp
```diff
@@ -161,6 +163,36 @@ int main(int argc, char** argv)
...
+    description_read_repetitions = JointStateListener::default_description_read_repetitions_;
+  }
+  double description_read_delay;
+  node.param("description_read_delay", description_read_delay, JointStateListener::default_description_read_delay_);
+  if (description_read_delay <= 0) {
+    ROS_WARN_STREAM("Parameter description_read_delay is smaller than or equal 0. Default value of " <<
+                    JointStateListener::default_description_read_delay_ << " seconds will be used.");
+    description_read_delay = JointStateListener::default_description_read_delay_;
+  }
+  for (size_t i = 0; i < description_read_repetitions; ++i) {
```
sloretz, line 184: Please add a test for this code.

sloretz (changes requested):
Just a couple changes

sloretz:
@clalancette Thoughts?

sgillen:
Sorry to nose in, are there plans to move ahead with this pull request? It would be useful for my application as well. I can help with the requested changes if need be.

clalancette:
I'm wondering if there is a nicer way to do this ,rather than just polling for a set amount of time.  That is, it seems to me like it might be better for robot_state_publisher to have a configurable topic that would get published to once the robot description is available.  That way, it would just wait around until that event happened, rather than polling the parameter server.  Thoughts about an approach like that?

ahcorde:
This PR is targeting an EOL ROS distro, closing this. Feel free to reopen this PR in an active ROS distro if the PR is relevant

PR #54, opened 2016-06-10
Add target_link_libraries for joint_state_listener library + install it

kartikmohta:
I assume it should be installed to let others use the library, but might be mistaken then.


jackieokay:
thanks, yeah, I added JointStateListener as a separate shared library recently but forgot to install it.

PR #55, opened 2016-07-05
Jade devel

cpeavy2:


jackieokay:
This doesn't appear to be a valid change, you are merging jade-devel into kinetic-devel. Those branches are supposed to remain separate to avoid introducing breaking changes from Kinetic into Jade. Let me know if you have any questions about the procedure to submit a pull request to this repository.

PR #60, opened 2016-09-05
Use urdf::*ShredPtr instead of boost::shared_ptr

jspricke:
urdfdom_headers uses C++ std::shared_ptr. As it exports it as custom
*SharedPtr type, we can use the to sty compatible.

This also adds a proper dependency for urdfdom-headers

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,14 +7,16 @@ find_package(catkin REQUIRED
 )
 find_package(Eigen3 REQUIRED)
 
+find_package(urdfdom_headers 0.4 REQUIRED)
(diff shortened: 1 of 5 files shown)


wjwwood:
I believe this would break Wily and Jessie, as it did for other repositories. I think it will need to use the compatibility header from `urdf` as we did in moveit and rviz.

jspricke:
@wjwwood ping.

wjwwood:
I still haven't tested this on Wily or Jessie, but I think edaf038 was meant to address this. I'll do a release and see what happens based on that.

PR #61, opened 2016-09-13
fix segfault on undefined robot_description

rhaschke:


src/joint_state_listener.cpp
```diff
@@ -152,7 +152,9 @@ int main(int argc, char** argv)
 
   // gets the location of the robot description on the parameter server
   urdf::Model model;
-  model.initParam("robot_description");
```
rhaschke, line 155: If initialization of model failed, `initParam()` returns false, which was ignored here.

jackieokay:
The warning in the automated build will be fixed with #62, so this looks good to me.

PR #64, opened 2016-10-03
remove undeserved error log for empty JointState.position

sbarthelemy:
as per the JointState message doc, the `position` member must be
either empty OR have the same size as the `name` member.

Some background here http://answers.ros.org/question/244796/jointstate-messages-with-heterogeneous-position-and-velocity-sensors/

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -88,7 +88,16 @@ void JointStateListener::callbackFixedJoint(const ros::TimerEvent& e)
 void JointStateListener::callbackJointState(const JointStateConstPtr& state)
 {
   if (state->name.size() != state->position.size()){
-    ROS_ERROR("Robot state publisher received an invalid joint state vector");
+    if (state->position.empty()){
+      const int throttleSeconds = 300;
+      ROS_WARN_THROTTLE(throttleSeconds,
+          "Robot state publisher ignored a JointState message about joint(s) "
+          "\"%s\"(,...) whose position member was empty. This message will "
+          "not reappear for %d seconds.", state->name[0].c_str(),
+          throttleSeconds);
+    } else {
+      ROS_ERROR("Robot state publisher ignored an invalid JointState message");
+    }
     return;
   }
 


wjwwood:
@tfoote what do you think about this?

tfoote:
The raw joint state message without position is not invalid in of itself, but I don't think that the robot_state_publisher can do anything without the position message. I think that publishing the two different Joint State messages is the best thing to do as it encapsulates the raw data. And you're right that the error in the joint_state_listener is overly strong. However I think that the code should probably be downgraded to a warning, and the message clarified that the message cannot be handled, instead of saying that it's invalid. 

Note that if this is downgraded to a warning (it should probably be throttled too) it should still ignore the message that it cannot publish.

sbarthelemy:
> I think that publishing the two different Joint State messages is the best thing to do as it encapsulates the raw data. 

In the end (and since I'm sticking to indigo right now), I circumvented the issue by publishing two JointState messages, in two topics ("joint_states" and "wheel_states"). robot_state_publisher only listens to "joint_states".

But I'm still interested in fixing this erroneous spammy log.

> The raw joint state message without position is not invalid in of itself, but I don't think that the robot_state_publisher can do anything without the position message. (...) However I think that the code should probably be downgraded to a warning, (...)

Well I personally don't see a use for this warning and it complicates the matter a bit since we'll need to choose a throttle period.

Similarly, there is no warning for empty JointState messages (not names nor position) which are also both valid and useless (ie. not handled by robot_state_publisher), and I don't see why there should be any.

But anyway, both solutions (no log or a throttled warning) would be an improvement, and I'll be happy to submit the one you prefer.

I just pushed the second one for review.

src/joint_state_listener.cpp
```diff
@@ -86,7 +86,11 @@ void JointStateListener::callbackFixedJoint(const ros::TimerEvent& e)
 void JointStateListener::callbackJointState(const JointStateConstPtr& state)
 {
   if (state->name.size() != state->position.size()){
-    ROS_ERROR("Robot state publisher received an invalid joint state vector");
+    if (state->position.empty()){
+      ROS_WARN_THROTTLE(300, "Robot state publisher ignored a JointState message whose position member was empty");
```
tfoote: It would be great to list the names, and also mention that this message will not reappear for 300 seconds. For those who don't know that this is a throttled output.

tfoote (commented):
Thanks for the PR I just had a two suggestions to improve the debuggability for new users.

sbarthelemy:
@tfoote: Hi, I updated the PR following your suggestions, could you have a look? Thx!

tfoote (approved):
Thanks that's nice and clear now.

PR #65, opened 2016-10-20
adding a parameter "ignore_timestamp"

facontidavide:
Hi,

very often I found the use case of "old" /joint_states published from rosbags, ignored by robot_state_publisher (by design, I know).
I have the feeling that,  instead of remapping the timestamp to the present (as it is usually done), it would be more intuitive to allow robot_state_publisher to just use the last JointState received, ignoring its timestamp.

The default behaviour is the same as it was before, so users should not find any backward-compatibility problem.

Davide

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -74,7 +74,8 @@ protected:
   ros::Time last_callback_time_;
   std::map<std::string, ros::Time> last_publish_time_;
   MimicMap mimic_;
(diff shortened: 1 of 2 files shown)


wjwwood (approved):
I made a small style change, but otherwise it lgtm.

wjwwood:
Once I merge this, can you please update the documentation on the wiki to document this new parameter: http://wiki.ros.org/robot_state_publisher

facontidavide:
I updated the wiki. Thanks for merging

PR #66, opened 2016-11-02
Make rostest dependency conditional, depending on CATKIN_ENABLE_TESTING

### LINKED ISSUE #58: Mandatory dependency on rostest
Hi,

I have noticed that the dependency on rostest is always there. In other packages ([TF example](https://github.com/ros/geometry/blob/indigo-devel/tf/CMakeLists.txt#L51)) rostest is only find-packaged if `CATKIN_ENABLE_TESTING` is true. [Reference issue](https://github.com/ros/rosdistro/issues/3010)

I wonder whether there is a reason for not doing the same in this package.

jvgomez:
Closes #58

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ target_link_libraries(state_publisher ${PROJECT_NAME}_solver ${orocos_kdl_LIBRAR
 
 if (CATKIN_ENABLE_TESTING)
 
-  find_package(rostest)
+  find_package(rostest REQUIRED)
 
   add_rostest_gtest(test_one_link ${CMAKE_CURRENT_SOURCE_DIR}/test/test_one_link.launch test/test_one_link.cpp)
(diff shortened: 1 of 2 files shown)


tfoote (approved):
lgtm

CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ target_link_libraries(state_publisher ${PROJECT_NAME}_solver ${orocos_kdl_LIBRAR
 
 if (CATKIN_ENABLE_TESTING)
 
+  find_package(rostest)
```
wjwwood: Please use `REQUIRED`.

wjwwood (changes requested):
One small change, otherwise, lgtm.

jvgomez:
@wjwwood I have done the requested changes.

wjwwood (approved):

PR #78, opened 2017-07-05
Added c++11 target_compile_options

sloretz:
This makes some targets use `std=c++11`. These files include headers from `urdfdom_headers`, which uses c++11 on some platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 project(robot_state_publisher)
 
 find_package(orocos_kdl REQUIRED)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -26,13 +26,16 @@ target_link_libraries(${PROJECT_NAME}_solver ${catkin_LIBRARIES} ${orocos_kdl_LI
 
 add_library(joint_state_listener src/joint_state_listener.cpp)
 target_link_libraries(joint_state_listener ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
+target_compile_options(joint_state_listener PUBLIC -std=c++11)
```
clalancette: Instead of doing this, what if we just hoisted the

```
  if(NOT WIN32)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
  endif()
```

from below and made it apply to the whole package?  I think this should be fine on all platforms that kinetic supports, right?

sloretz: Looks good to me.

If I understand correctly, `target_compile_options()` with `PUBLIC` means the option propagates to targets in other projects that depend or link against `joint_state_listener` as long as the target from this project is exported. However I don't see any c++11 features in the api on the kinetic-devel branch, so It doesn't seem like the c++11 option needs to be exported. Also I don't see anything exporting/installing CMake targets. Maybe catkin exports the targets automatically from the `catkin_package` macro? Regardless `joint_state_listener` isn't listed when that macro is called.

clalancette:
@sloretz I did an update based on what we discussed earlier.  Take a look when you get a chance.

sloretz:
@clalancette ah you tricked me! I tried to approve, but I can't approve my own pull request. I'll squash/merge it instead.

clalancette:
Heh, nice :).  That does seem to be a downside to the github model; it's odd about what you do for reviews/approvals on collaborative branches.  Anyway, thanks for taking a look and merging.

Issue #79, opened 2017-08-31
Remove dependency on tf


sloretz:
It looks like robot_state_publisher already uses tf2 for publishing transforms, but depends on `tf` so it can use `tf::resolve` on a frame given a tf_prefix. Removing the tf dependency would required dropping support for tf_prefix ([which was removed in tf2](http://wiki.ros.org/tf2/Migration#tf_and_tf2)), and [would break ABI compatibility](https://github.com/ros/robot_state_publisher/blob/530bba1d7001afb2d61dea6c52b966111f927f20/include/robot_state_publisher/robot_state_publisher.h#L79) so it should be done no sooner than ROS Melodic.

andrestoga:
Hi,

I have a question: If robot_state_publisher uses tf2 for publishing transforms why is the topic message type /tf still tf/tfMessage instead of tf2_msgs/TFMessage?

best,

sloretz:
Hi @andrestoga, You'll probably get a better answer on https://answers.ros.org or on https://github.com/ros/geometry2 .

It may have something to do with the two message types being identical:
* [tf2 message](https://github.com/ros/geometry2/blob/indigo-devel/tf2_msgs/msg/TFMessage.msg)
* [tf1 message](https://github.com/ros/geometry/blob/indigo-devel/tf/msg/tfMessage.msg)

sloretz:
Fixed by #82

PR #82, opened 2018-02-13
Remove dependency on tf and tf_prefix support

sloretz:
This PR removes the dependency on tf (#79) and the long deprecated tf_prefix. I have not done a prerelease test yet. It is likely the removal of the `tf_prefix` argument will break downstream users of the c++ API.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
 
 find_package(orocos_kdl REQUIRED)
 find_package(catkin REQUIRED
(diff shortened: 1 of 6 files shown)


clalancette (approved):
The code looks OK to me.  Please do a prerelease on the downstream users and open up PRs for the broken ones before merging, though.

sloretz:
Looks like a `--level 2` prerelease test successfully builds with just the 2 referenced PRs. There were a few packages whose apt dependencies couldn't be installed: `magni_demos`,  `magni_robot`,  `minas`,  `tra1_bringup`,  `tra1_description`,  `tra1_moveit_config`. However, all of these run_depend or exec_depend on this package so I don't expect them to be affected. I think it is safe to merge this.

PR #83, opened 2018-02-15
added warning when joint is found in joint message but not in the urdf

jgueldenstein:
when the joint message includes a joint which is not found in the URDF, a warning is now displayed. Previously this was only ignored.

Since no transform can be published when no URDF description is provided for the joint, either the URDF is incomplete or the `joint_state_controller` publishes wrong joints.

@SammyRamone
@v4hn


sloretz (approved):
Thanks for the pull request!

clalancette (approved):

PR #84, opened 2018-02-15
added ros_warn if JointState message is older than 30 seconds

jgueldenstein:
This pr adds a throttled warning when the joint state message is older than 30 seconds.

Since transforms should be up to date I think it is useful to notify about the joint state being outdated. This can help with debugging when the `joint_state_controller` sends wrong timestamps due to out of sync clocks or uninitialized headers.

@SammyRamone
@v4hn


sloretz (approved):
Thanks for the pull request!

clalancette (approved):

PR #87, opened 2018-02-26
Add in a public default constructor

### LINKED ISSUE #23: no public default constructor.

clalancette:
This should fix #23, and obsoletes #43 .  This PR does a number of things:

1.  It pushes the remaining TF dependency as deeply as possible.  To completely get rid of TF would mean we would have to get rid of the `tf_prefix` functionality.
1.  It moves the include of `boost::shared_ptr` over to the one place it is needed.  We basically can't get rid of this use of boost, since it is what is returned in the callback.
1.  It does some cleanup of namespaces.
1.  It switches the package to package format 2.
1.  It removes the deprecated name of `robot_state_publisher` (`state_publisher`).  That's been deprecated since a commit in 2012, so it is probably safe to remove.
1.  It splits the `main` out from the `joint_state_listener.cpp` file and into its own file.  This makes it so that `JointStateListener` could be used as a library on its own now.
1.  It adds in public default constructors for both `RobotStatePublisher` and `JointStateListener`.

It is easiest to review this PR on a commit-by-commit basis, since the individual commits all make sense.  Let me know if you'd prefer me to split this up into separate PRs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,10 +51,6 @@ target_link_libraries(joint_state_listener ${PROJECT_NAME}_solver ${orocos_kdl_L
 add_executable(${PROJECT_NAME} src/joint_state_listener.cpp)
 target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
 
-# compile the same executable using the old name as well
-add_executable(state_publisher src/joint_state_listener.cpp)
-target_link_libraries(state_publisher ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
-
 # Tests
 
 if (CATKIN_ENABLE_TESTING)
@@ -95,7 +91,7 @@ install(TARGETS ${PROJECT_NAME}_solver joint_state_listener
   RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
 
 # Install executable
-install(TARGETS ${PROJECT_NAME} state_publisher
+install(TARGETS ${PROJECT_NAME}
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
 install(DIRECTORY include/${PROJECT_NAME}/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,8 +48,8 @@ target_link_libraries(${PROJECT_NAME}_solver ${catkin_LIBRARIES} ${orocos_kdl_LI
 add_library(joint_state_listener src/joint_state_listener.cpp)
 target_link_libraries(joint_state_listener ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
 
-add_executable(${PROJECT_NAME} src/joint_state_listener.cpp)
-target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
+add_executable(${PROJECT_NAME} src/robot_state_publisher_node.cpp)
+target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_solver joint_state_listener ${orocos_kdl_LIBRARIES})
 
 # Tests
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,9 @@ if (CATKIN_ENABLE_TESTING)
 
   find_package(rostest REQUIRED)
 
+  add_rostest_gtest(test_default_constructor test/test_default_constructor.launch test/test_default_constructor.cpp)
+  target_link_libraries(test_default_constructor joint_state_listener ${PROJECT_NAME}_solver)
+
   add_rostest_gtest(test_one_link ${CMAKE_CURRENT_SOURCE_DIR}/test/test_one_link.launch test/test_one_link.cpp)
   target_link_libraries(test_one_link ${catkin_LIBRARIES} ${PROJECT_NAME}_solver)
 
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -56,6 +56,10 @@ namespace robot_state_publisher {
 
 class JointStateListener {
 public:
+  /** Default constructor.
+   */
+  JointStateListener();
+
   /** Constructor
    * \param tree The kinematic model of a robot, represented by a KDL Tree
    */
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -38,6 +38,8 @@
 #define JOINT_STATE_LISTENER_H
 
 #include <memory>
+#include <map>
+#include <string>
 
 #include <boost/scoped_ptr.hpp>
 #include <urdf/model.h>
@@ -46,10 +48,6 @@
 #include <sensor_msgs/JointState.h>
 
 #include "robot_state_publisher/robot_state_publisher.h"
-
-using namespace std;
-using namespace ros;
-using namespace KDL;
 
 typedef boost::shared_ptr<sensor_msgs::JointState const> JointStateConstPtr;
 typedef std::map<std::string, urdf::JointMimicSharedPtr > MimicMap;
@@ -73,9 +71,9 @@ protected:
   virtual void callbackJointState(const JointStateConstPtr& state);
   virtual void callbackFixedJoint(const ros::TimerEvent& e);
 
-  Duration publish_interval_;
+  ros::Duration publish_interval_;
   std::shared_ptr<RobotStatePublisher> state_publisher_;
-  Subscriber joint_state_sub_;
+  ros::Subscriber joint_state_sub_;
   ros::Timer timer_;
   ros::Time last_callback_time_;
   std::map<std::string, ros::Time> last_publish_time_;
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -49,10 +49,10 @@
 
 #include "robot_state_publisher/robot_state_publisher.h"
 
+namespace robot_state_publisher {
+
 typedef boost::shared_ptr<sensor_msgs::JointState const> JointStateConstPtr;
 typedef std::map<std::string, urdf::JointMimicSharedPtr > MimicMap;
-
-namespace robot_state_publisher {
 
 class JointStateListener {
 public:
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -29,12 +29,8 @@ target_link_libraries(${PROJECT_NAME}_solver ${catkin_LIBRARIES} ${orocos_kdl_LI
 add_library(joint_state_listener src/joint_state_listener.cpp)
 target_link_libraries(joint_state_listener ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
 
-add_executable(${PROJECT_NAME} src/joint_state_listener.cpp)
-target_link_libraries(${PROJECT_NAME} ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
-
-# compile the same executable using the old name as well
-add_executable(state_publisher src/joint_state_listener.cpp)
-target_link_libraries(state_publisher ${PROJECT_NAME}_solver ${orocos_kdl_LIBRARIES})
```
sloretz, line 56: I think removing `state_publisher` should be done in `melodic-devel`. Kinetic and lunar have releases where this name is available, so it's possible there are launch files on robots using it.

include/robot_state_publisher/robot_state_publisher.h
```diff
@@ -37,9 +37,10 @@
 #ifndef ROBOT_STATE_PUBLISHER_H
 #define ROBOT_STATE_PUBLISHER_H
 
+#include <map>
+#include <string>
+
 #include <ros/ros.h>
-#include <boost/scoped_ptr.hpp>
-#include <tf/tf.h>
```
sloretz: Removing `<tf/tf.h>` in #82 on `melodic-devel` required ros-naoqi/naoqi_driver#103 . Maybe it should be left in place to avoid breaking source builds of other kinetic/lunar packages?

include/robot_state_publisher/joint_state_listener.h
```diff
@@ -37,24 +37,29 @@
...
 typedef boost::shared_ptr<sensor_msgs::JointState const> JointStateConstPtr;
 typedef std::map<std::string, urdf::JointMimicSharedPtr > MimicMap;
 
 namespace robot_state_publisher {
 
 class JointStateListener {
 public:
+  /** Default constructor.
+   */
+  JointStateListener();
```
sloretz, line 61: I haven't looked through the member types, but does JointStateListener meet the requirements to get a default copy or move assignment operator? If not, then I don't see a way to go from a default constructed `JointStateListener` to a functional one. Mind adding a unit test to check if the assignment operator works?

include/robot_state_publisher/robot_state_publisher.h
```diff
@@ -63,6 +64,10 @@ class SegmentPair
 class RobotStatePublisher
 {
 public:
+  /** Default constructor.
+   */
+  RobotStatePublisher();
```
sloretz, line 69: Same comment about assignment operator

include/robot_state_publisher/robot_state_publisher.h
sloretz: Targeting noetic

CMakeLists.txt
sloretz: targeting noetic

include/robot_state_publisher/joint_state_listener.h
sloretz: Made sure `JointStateListener` move assignment works in bc0cb46

include/robot_state_publisher/robot_state_publisher.h
sloretz: Fixed RobotStatePublisher move assignment in bc0cb46

sloretz (approved):
LGTM! @clalancette I rebased to fix conflicts and added a couple things to resolve old comments. Mind taking a look?

package.xml
```diff
@@ -26,28 +26,21 @@
...
-  <run_depend>rosconsole</run_depend>
-  <run_depend>roscpp</run_depend>
-  <run_depend>rostime</run_depend>
-  <run_depend>sensor_msgs</run_depend>
-  <run_depend>tf2_ros</run_depend>
-  <run_depend>tf2_kdl</run_depend>
+  <build_depend>eigen</build_depend>
 
-  <test_depend>rosbag</test_depend>
+  <exec_depend>liborocos-kdl-dev</exec_depend>
```
clalancette: Shouldn't this be:

`<exec_depend>liborocos-kdl</exec_depend>`

src/joint_state_listener.cpp
```diff
@@ -66,7 +70,7 @@ JointStateListener::JointStateListener(const std::shared_ptr<RobotStatePublisher
   // ignore_timestamp_ == true, joins_states messages are accepted, no matter their timestamp
   n_tilde.param("ignore_timestamp", ignore_timestamp_, false);
   // get the tf_prefix parameter from the closest namespace
-  publish_interval_ = ros::Duration(1.0/max(publish_freq, 1.0));
+  publish_interval_ = ros::Duration(1.0/std::max(publish_freq, 1.0));
```
clalancette, line 74: Nit: `#include <algorithm>` at the top of the file for `std::max`.

src/joint_state_listener.cpp
```diff
@@ -123,7 +127,7 @@ void JointStateListener::callbackJointState(const JointStateConstPtr& state)
 
   // determine least recently published joint
   ros::Time last_published = now;
-  for (unsigned int i=0; i<state->name.size(); i++) {
+  for (unsigned int i = 0; i < state->name.size(); i++) {
```
clalancette: Since we are changing this line, we should probably just make i a `size_t` to be most correct.

src/joint_state_listener.cpp
```diff
@@ -133,8 +137,8 @@ void JointStateListener::callbackJointState(const JointStateConstPtr& state)
   // check if we need to publish
   if (ignore_timestamp_ || state->header.stamp >= last_published + publish_interval_) {
     // get joint positions from state message
-    map<string, double> joint_positions;
-    for (unsigned int i=0; i<state->name.size(); i++) {
+    std::map<std::string, double> joint_positions;
+    for (unsigned int i = 0; i < state->name.size(); i++) {
```
clalancette: Same here; make i a `size_t`.

src/robot_state_publisher.cpp
```diff
@@ -58,7 +62,7 @@ void RobotStatePublisher::addChildren(const KDL::SegmentMap::const_iterator segm
   const std::string& root = GetTreeElementSegment(segment->second).getName();
 
   const std::vector<KDL::SegmentMap::const_iterator>& children = GetTreeElementChildren(segment->second);
-  for (unsigned int i=0; i<children.size(); i++) {
+  for (unsigned int i = 0; i < children.size(); ++i) {
```
clalancette: `size_t` for i.

src/robot_state_publisher_node.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  KDL::Tree tree;
+  if (!kdl_parser::treeFromUrdfModel(model, tree)) {
+    ROS_ERROR("Failed to extract kdl tree from xml robot description");
+    return 1;
+  }
+
+  robot_state_publisher::MimicMap mimic;
+
+  for(std::map< std::string, urdf::JointSharedPtr >::iterator i = model.joints_.begin(); i != model.joints_.end(); i++) {
```
clalancette:
```suggestion
  for (std::map< std::string, urdf::JointSharedPtr >::iterator i = model.joints_.begin(); i != model.joints_.end(); i++) {
```

src/robot_state_publisher_node.cpp
```diff
@@ -0,0 +1,78 @@
...
+    if(i->second->mimic) {
```
clalancette:
```suggestion
    if (i->second->mimic) {
```

src/robot_state_publisher_node.cpp
```diff
@@ -0,0 +1,78 @@
...
+      mimic.insert(make_pair(i->first, i->second->mimic));
```
clalancette:
```suggestion
      mimic.insert(std::make_pair(i->first, i->second->mimic));
```

And `#include <utility>` at the top.

test/test_default_constructor.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <gtest/gtest.h>
+#include <ros/ros.h>
+
+#include "robot_state_publisher/joint_state_listener.h"
+
+
+TEST(RobotStatePublisher, default_constructor)
+{
+  robot_state_publisher::RobotStatePublisher();
```
clalancette: This is really only testing that we don't throw an exception in the constructor.  I think this might be better if we sub-class `RobotStatePublisher` in the test, then do some standard `EXPECT` checks on the `protected` member variables of the class.  What do you think?

clalancette (commented):
A bunch of mostly minor things to fix, but I think this is close!

src/robot_state_publisher_node.cpp
sloretz: Include `<utility>` in 8aedf9305b85d0d6a0b9437956b955395a5c0e2f

src/robot_state_publisher.cpp
sloretz: size_t in d780506c444f1519c78c29717b7f4296a884b016

src/joint_state_listener.cpp
sloretz: size_t in d780506c444f1519c78c29717b7f4296a884b016

src/joint_state_listener.cpp
sloretz: size_t in d780506c444f1519c78c29717b7f4296a884b016

src/joint_state_listener.cpp
sloretz: Include <algorithm> in 0826b6aae67b321fc50cb44d50679a9b28247def

package.xml
sloretz: `<build_export_depend>liborocos-kdl-dev` because headers are included in headers we install, and `<exec_depend>liborocos-kdl` in da6f692513c924a939946fa13dce3b017dc6a41f

test/test_default_constructor.cpp
sloretz: on second thought I'm not sure the default constructor test added value since the default constructor is used in the assignment tests. Checking protected variables seems too low level since we have higher level tests checking behavior. I deleted the `default_constructor` tests in 361870e9a3f9dba082b493177a8444fa293b1488

clalancette (commented):
This looks good to me, thanks for iterating @sloretz .  GitHub says "Pull request authors can't approve their own pull request", so please take this comment as approval.

PR #89, opened 2018-07-03
Check for private robot_description

oKermorgant:
Hi, 

I came across a use case where I want robot_state_publisher to use its own version of URDF (namely: without meshes as they are loaded with absolute pathes which can lead to unfound mesh files when using several robots).
A solution is to run robot_state_publisher in its own namespace, another is to allow a private robot_description parameter to be used.

If you find any interest in this PR then feel free to merge, otherwise I'll use my fork on our system.

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -170,9 +170,13 @@ int main(int argc, char** argv)
   ///////////////////////////////////////// end deprecation warning
 
   // gets the location of the robot description on the parameter server
+  // check private then namespace parameter
   urdf::Model model;
-  if (!model.initParam("robot_description"))
-    return -1;
+  if (!model.initParamWithNodeHandle("robot_description", NodeHandle("~")))
+  {
+    if (!model.initParam("robot_description"))
+      return -1;
+  }
 
   KDL::Tree tree;
   if (!kdl_parser::treeFromUrdfModel(model, tree)) {


sloretz:
Thanks for the PR @oKermorgant

I think remapping can already do this. If you're using roslaunch then adding  `<remap from="robot_description" to="~/robot_description"/>` should make `robot_state_publisher` use a private parameter.

oKermorgant:
Hi @sloretz , 

I tried remapping and although this kind of global / private remapping should be ok, it seems it is not working with urdf::Model::initParam.

I guess it is more on the urdf side than robot_state_publisher, I'll thus remap robot_description to a new name (but still global), which works.

sloretz:
It looks like `robot_description` can be remapped to a private name as long as `from` is `/robot_description`. I don't know why `from="robot_description"` doesn't seem to work. @oKermorgant would this workaround be fine for your use-case?

Here is a launch file that works on melodic

```xml
<?xml version="1.0"?>
<launch>
  <param name="robot_description" type="str" value="&lt;robot name=&quot;global_robot&quot;&gt;&lt;link name=&quot;global_link&quot;/&gt;&lt;/robot&gt;"/>
  <node pkg="robot_state_publisher" name="global_robot_state_pub" type="robot_state_publisher"/>

  <node pkg="robot_state_publisher" name="private_robot_state_pub" type="robot_state_publisher">
    <remap from="/robot_description" to="~/robot_description"/>
    <param name="~/robot_description" type="str" value="&lt;robot name=&quot;private_robot&quot;&gt;&lt;link name=&quot;private_link&quot;/&gt;&lt;/robot&gt;"/>
  </node>
</launch>
```

oKermorgant:
@sloretz thanks for the time, it is still not my use case as I use robot namespaces.
I can manage without an actual private parameter:
```
<group ns="auv1">
    <param name="robot_description" command="$(find xacro)/xacro  $(find ecn_bluerov)/urdf/pirov_full.xacro"/>
    <param name="robot_description_no_mesh" command="$(find xacro)/xacro $(find ecn_bluerov)/urdf/pirov_rdx.xacro"/>
    
    <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher">
        <remap from="robot_description" to="robot_description_no_mesh"/>
    </node>
</group>
```
From the urdf::initParam code I narrowed this error behavior to [the use of nh.searchParam](https://github.com/ros/urdf/blob/470729633ba0419735603a054fec0b5bbfe442f9/urdf/src/model.cpp#L94) that does not work well on private parameters.
Anyway as shown in my example we can manage without private parameters and my PR it probably too much targeted to this particular use.

sloretz:
Thanks for narrowing it down @oKermorgant . I opened ros/urdf#20, and will close this PR for now.

PR #92, opened 2019-01-04
Fixed problem when building static library version

ivanpauno:
ARCHIVE DESTINATION wasn't set in cmake install directive, so static building failed.
This solves the issue.


sloretz (approved):

sloretz:
Thanks for the PR!

PR #97, opened 2019-01-15
Handle empty URDF file. Fixes #81

### LINKED ISSUE #81: robot-state-publisher crashes on an empty URDF
Create a launch file with the following content. It initializes `robot_description` to contain a urdf that has no links.

```
<?xml version="1.0"?>
<launch>
  <param name="robot_description" value='&lt;robot name="foo" xmlns:xacro="http://ros.org/wiki/xacro"&gt;&lt;/robot&gt;'/>
  <node pkg="robot_state_publisher" name="robot_state_publisher" type="robot_state_publisher" output="screen"/>
</launch>
```

`robot_state_publsiher` crashes with the following output.

```
process[robot_state_publisher-2]: started with pid [316]
[ERROR] [1518481302.405483470]: No link elements found in urdf file
[robot_state_publisher-2] process has died [pid 316, exit code -11, cmd /opt/ros/indigo/lib/robot_state_publisher/robot_state_publisher __name:=robot_state_publisher __log:=/home/developer/.ros/log/dd4738a0-1053-11e8-81e1-0242ac110005/robot_state_publisher-2.log].
log file: /home/developer/.ros/log/dd4738a0-1053-11e8-81e1-0242ac110005/robot_state_publisher-2*.log
```

The node should exit in this case since there's nothing it can do with an empty URDF, but it shouldn't crash.

betab0t:
Node will exit cleanly if failed to load URDF **in case it has no links**. Fixes #81

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -171,8 +171,12 @@ int main(int argc, char** argv)
 
   // gets the location of the robot description on the parameter server
   urdf::Model model;
-  if (!model.initParam("robot_description"))
+  if (!model.initParam("robot_description")) {
+    if (model.links_.empty()) {
+      return 0;
+    }
     return -1;
+  }
 
   KDL::Tree tree;
   if (!kdl_parser::treeFromUrdfModel(model, tree)) {
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -175,13 +175,13 @@ int main(int argc, char** argv)
     if (model.links_.empty()) {
       return 0;
     }
-    return -1;
+    return 1;
   }
 
   KDL::Tree tree;
   if (!kdl_parser::treeFromUrdfModel(model, tree)) {
     ROS_ERROR("Failed to extract kdl tree from xml robot description");
-    return -1;
+    return 1;
   }
 
   MimicMap mimic;


src/joint_state_listener.cpp
```diff
@@ -171,8 +171,12 @@ int main(int argc, char** argv)
 
   // gets the location of the robot description on the parameter server
   urdf::Model model;
-  if (!model.initParam("robot_description"))
+  if (!model.initParam("robot_description")) {
+    if (model.links_.empty()) {
+      return 0;
+    }
     return -1;
```
clalancette: I know you didn't add this, but -1 (really 255 in bash) is considered a reserved exit code.  Replacing this with 1 would be better.

src/joint_state_listener.cpp
```diff
@@ -171,8 +171,12 @@ int main(int argc, char** argv)
 
   // gets the location of the robot description on the parameter server
   urdf::Model model;
-  if (!model.initParam("robot_description"))
+  if (!model.initParam("robot_description")) {
+    if (model.links_.empty()) {
+      return 0;
```
clalancette, line 176: I'm thinking that we don't want this to be successful; it seems to me that launching robot_state_publisher with a URDF that has no links should probably be louder (use `ROS_ERROR` to print a message), and return an error code (1 seems appropriate here).

src/joint_state_listener.cpp
```diff
@@ -171,8 +171,12 @@ int main(int argc, char** argv)
 
   // gets the location of the robot description on the parameter server
   urdf::Model model;
-  if (!model.initParam("robot_description"))
+  if (!model.initParam("robot_description")) {
+    if (model.links_.empty()) {
```
clalancette, line 175: This seems to be in a weird location; we are only checking if the links are empty if we *failed* to initialize the URDF.  But in that case, we are going to quit anyway since we can't do anything.  Doesn't this check want to be down below, after a successful `initParam`?

clalancette (changes requested):

src/joint_state_listener.cpp
betab0t: Hi @clalancette and thank you for the super quick review :smile: 
I'll explain: [initParam](https://github.com/ros/urdf/blob/470729633ba0419735603a054fec0b5bbfe442f9/urdf/src/model.cpp#L85) will fail on urdf without any links because it calles [initString](https://github.com/ros/urdf/blob/470729633ba0419735603a054fec0b5bbfe442f9/urdf/src/model.cpp#L119) which in turn call [parseURDF](https://github.com/ros/urdf/blob/470729633ba0419735603a054fec0b5bbfe442f9/urdf/src/model.cpp#L201)... [which will fail and print an error if no links](https://github.com/ros/urdfdom/blob/06f5f9bc34f09b530d9f3743cb0516934625da54/urdf_parser/src/model.cpp#L178). so adding the check after it will have no effect.

src/joint_state_listener.cpp
betab0t: As mentioned in #81 ROS_ERROR is [already printed by urdfdom](https://github.com/ros/urdfdom/blob/master/urdf_parser/src/model.cpp#L179)

src/joint_state_listener.cpp
betab0t: https://github.com/ros/robot_state_publisher/blob/ff466b4ffec2e352664822a830ba795480d4edcf/src/joint_state_listener.cpp#L180
done :+1: also noticed line 180 is returning -1 so i modified it to match...

src/joint_state_listener.cpp
clalancette: Hm, OK.  I guess I'm a bit confused about this PR then.  Before this PR, the code looked like this:

```
  if (!model.initParam("robot_description"))
    return -1;
```

Thus, if `parseURDF` returned a nullptr, which caused `initParam` to return false, wouldn't we *already* be avoiding a segmentation fault?  That is, what is the difference that this PR make?

betab0t: Maybe I should have better explained it, you are right we are already avoiding segfault, the goal of this PR is only to make sure it won't look as if we segfaulted, by only printing ROS_ERROR regarding the missing links and that's it, exit cleanly(0).

clalancette: Hm, so the reason I'm so confused is that in the original report, @sloretz was seeing a segmentation fault (that's the -11 error code he was originally reporting).  But with the current robot_state_publisher (without this PR at all), I can't reproduce that; it just exits with:

```
[ERROR] [1547660920.315867675]: No link elements found in urdf file
[robot_state_publisher-2] process has died [pid 12998, exit code 255, cmd /home/ubuntu/melodic/catkin_ws/install_isolated/lib/robot_state_publisher/robot_state_publisher __name:=robot_state_publisher __log:=/home/ubuntu/.ros/log/f4c98e34-19b6-11e9-9480-00163e5a3681/robot_state_publisher-2.log].
```

So now I'm wondering whether we need this PR (and this issue) at all; we are *already* checking that it is valid, and exiting.  The only thing this PR will change is the return code, which may be worthwhile, but is a lot smaller scope than the original.

@sloretz Any thoughts here?

sloretz: It looks like the ticket I filed has a path to `indigo`. Maybe I was using an outdated version of robot state publisher?

betab0t: @sloretz @clalancette yes, looks like it was fixed in https://github.com/ros/robot_state_publisher/commit/b9029f7d5830aceeabfdd94048fb4df4def2d9e0 and never backported to indigo

sloretz: @betab0t thanks for the PR and discovering that commit! I'll close the original issue, sorry for the confusion.

betab0t: @sloretz happy to help! @clalancette I'm closing this PR + I'll open a separate PR for the exit code.

clalancette: @betab0t That sounds great, it is much appreciated.

PR #98, opened 2019-01-17
Changed node returned error code since its considered reserved

betab0t:
Whenever robot_state_publisher node fails to parse URDF file it exits with error code -1/255 which is considered a reserved bash exit code(exit status out of range), changed to 1.


sloretz (approved):

sloretz:
Thanks for the PR!

PR #104, opened 2019-02-15
update how compiler flags are added

kejxu:
update how `c++11` compiler flag is added:
- only add it when it's supported (tested on Ubuntu 18.04)
- add similar logic for MSVC (as mentioned in the comment, this check would fail, but c++11 is always enabled). this pattern could be helpful if later this project moves to c++14 (which is an available switch in MSVC)

the other way to do this is to use `target_compile_features(mylib PUBLIC cxx_std_11)` from [cmake](https://cmake.org/cmake/help/v3.8/manual/cmake-compile-features.7.html#requiring-language-standards), but this would mean to add this line for every target. would this be preferred instead of adding a global `c++11` flag?

there shouldn't be too much difference between `add_compile_options` and `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} xxx")`, yet here are some discussions regarding the difference between the two: https://stackoverflow.com/questions/39501481/difference-between-add-compile-options-and-setcmake-cxx-flags. personally `add_compile_options` is preferred since it offers a clearer scope and better readability

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,15 @@
 cmake_minimum_required(VERSION 2.8)
 project(robot_state_publisher)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag("-std=c++11" COMPILER_SUPPORTS_CXX11)
+if(COMPILER_SUPPORTS_CXX11)
+  add_compile_options(-std=c++11)
+  if (NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")  
+  endif()
+endif()
+
 
 find_package(orocos_kdl REQUIRED)
 find_package(catkin REQUIRED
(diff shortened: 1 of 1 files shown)


sloretz:
Thanks for the PRs @kejxu. Same comment as https://github.com/ros/kdl_parser/pull/24#issuecomment-474175891, where personally I'd rather use the `CMAKE_CXX_STANDARD` variable.

@clalancette thoughts here too?

clalancette (approved):
I'll note that this subtly changes the link flags on Linux; before, we had `-std=c++11` there, and after this change, we do not.  That being said, I don't think that it actually has an effect, so I'm good with the change as-is.

kejxu:
> I'll note that this subtly changes the link flags on Linux;

Just curious, is this because of the difference between `add_compile_options` and `set(CMAKE_CXX_FLAGS ...)`?

clalancette:
> Thanks for the PRs @kejxu. Same comment as [ros/kdl_parser#24 (comment)](https://github.com/ros/kdl_parser/pull/24#issuecomment-474175891), where personally I'd rather use the `CMAKE_CXX_STANDARD` variable.
> 
> @clalancette thoughts here too?

Arg, I totally missed your comment here, sorry @sloretz .  I don't have a good feeling on which is preferred, honestly.  I'm also fine with switching this to C++14, though I don't think we are using any features which require it so I'm not too fussed either way.

@sloretz Let me know if you'd like me to make a change, I'm happy to follow up.

PR #106, opened 2019-03-05
Custom robot state publisher

peci1:
Last piece missing from #44 (for #25) to allow people using their own implementations of RobotStatePublisher.

This is definitely ABI and API break, however it might be worth it.


clalancette:
I'm not opposed to this change, though as you say it is an API and ABI break.  I'd really rather not do that for Kinetic, and I'm hesitant about doing it for Melodic.  That being said, I think few if any people are actively subclassing `RobotStatePublisher/JointStateListener` today, so the downstream breakage would probably be minimal.  @sloretz thoughts?

sloretz:
Given the ABI breakage, I'm fine with this change for Noetic.

peci1:
Okay. Should I retarget this PR to melodic-devel, or should I wait until a noetic-devel branch is created?

clalancette:
@peci1 It's best to wait for a noetic-devel branch.  I would create one right now, but I think @sloretz and I need to finish figuring out what to do about #108 before we do that.  Once we figure it out, then I'll create that branch for you.  Sorry for the wait.

peci1:
No problem.

clalancette:
@peci1 FYI, I finally created the `noetic-devel` branch and retargeted this PR to it.  It looks like you'll have to rebase here.  Thanks for the patience.

peci1:
See #115.

PR #109, opened 2019-03-22
make use_tf_static false by default

### LINKED ISSUE #105: use_tf_static should be false by default
The behaviour of publishing fixed joints changed significantly since the introduction of `use_tf_static`. By default, fixed joint transforms are published once to `tf_static` with the earliest possible timestamp that robot transformations can have.
When publishing joint positions and logging the robot state (as the set of its TFs) some time after the `robot_state_publisher` has been started, the resulting bagfile will contain static TFs from a time before the logging actually started and hence is not a correct representation of the robot state.

Although the static transforms are latched (and hence logged), they are not latched by default when replaying a rosbag. As a consequence it is not possible to start any node that needs the full robot state (e.g. RViz) after the replay started.

When reading the TFs into a TF buffer (which caches TFs for 10s by default), this creates distinct TF trees and results in failed lookups:
```
Lookup would require extrapolation at time 1551113242.324176550, but only time 1551109106.787312031 is in the buffer
Lookup would require extrapolation at time 1551113242.357692480, but only time 1551109106.787312031 is in the buffer
Lookup would require extrapolation into the future.  Requested time 1551113242.391208649 but the latest data is at time 1551113242.357692480
Lookup would require extrapolation into the future.  Requested time 1551113242.424724579 but the latest data is at time 1551113242.357692480
```
In this particular case the static transforms were published at 1551109106.787312031 and the non-static transforms were published at 1551113242.391208649.

By publishing fixed joints as static TF periodically together with the non-static TF, the `robot_state_publisher` would provide a complete state of the robot at any point in time and the before mentioned issues would be circumvented.

christian-rauch:
This PR reverts the behaviour of the static transform publisher and sets `use_tf_static` to false by default. The main reason for this is to fix the inconsistent behaviour when replaying latched static transforms.
When static transforms are published latched to `tf_static`, it is not possible to "joint" the stream of transformations at a later point.
E.g. it is not possible to recreate the complete state of a robot static transforms if replayed with  `rosbag play <file>.bag -s 10 -k` or when a transform listener is started after replaying the bagfile (see also: https://github.com/ros/robot_state_publisher/issues/105). Another issue is that the latching attribute is not preserved when post-processing a bagfile.

Additionally, this aligns with the settings in `RobotStatePublisher`:
https://github.com/ros/robot_state_publisher/blob/316ee45abe5a6bccf37d15e38b68eda99eb5158d/include/robot_state_publisher/robot_state_publisher.h#L78

Fixes #105 .

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -57,7 +57,7 @@ JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m,
   double publish_freq;
   n_tilde.param("publish_frequency", publish_freq, 50.0);
   // set whether to use the /tf_static latched static transform broadcaster
-  n_tilde.param("use_tf_static", use_tf_static_, true);
+  n_tilde.param("use_tf_static", use_tf_static_, false);
   // ignore_timestamp_ == true, joins_states messages are accepted, no matter their timestamp
   n_tilde.param("ignore_timestamp", ignore_timestamp_, false);
   // get the tf_prefix parameter from the closest namespace


sloretz:
Thanks for opening a PR. While the issues mentioned: only the last static publisher's message being latched, and the bug in rosbag's post-processing loosing latched status can be worked around with this setting, I disagree that it should be the default value. I'm still concerned about the extra bandwidth given that `/tf_static` usually has a lot of subscribers.

christian-rauch:
The critical scenario is the following:
Let's say a `robot_state_publisher` is started at t=0 and from thereon publishes static transforms latched with timestamp t=0. The robot driver becomes active at t=A and publishes joint states, which get transformed with timestamp t>A via the `robot_state_publisher`. If you start to record a rosbag a t=B, it will contain the timestamps with t=0 (static TFs) and t>B (dynamic TFs).

However, if you later play back the rosbag with a start timestamp of e.g. t=B+5 it will not publish the latched messages with t=0. This will prevent the tf buffer from generating a full robot kinematic tree.

I prefer correctness of the kinematic tree reconstruction at any point in time over bandwidth savings.

clalancette:
@sloretz and I discussed this further.  While what you are describing is a problem, it is essentially a problem in rosbag and/or the lower layers of the ROS 1 stack.  The same problem exists for *any* latched publisher, of which `tf` is only one.  Therefore, we don't think it is appropriate to work around this problem at the `robot_state_publisher` level.  Instead, we suggest pursuing a fix/behavior change for rosbag that would handle this in a more generic way.

Given the above, I'm going to close this out.  Feel free to keep commenting if you have other questions/comments.  Thanks for the contribution.

PR #112, opened 2019-08-14
Add joint_state_listener to the catkin package LIBRARIES

sevangelatos:
The joint_state_listener library was missing from the catkin_package libraries. As a consequence, one cannot to link against it from another package, without resorting to workarounds.

I just added it to the LIBRARIES section of the catkin_package macro


clalancette (approved):
Makes sense to me.

Issue #108, opened 2019-03-19
Release from melodic-devel branch


sloretz:
I was wondering why no PR job ran for #103, and it looks like it was released to melodic from the `kinetic-devel` branch

https://github.com/ros/rosdistro/blob/25b9046eda193ed47cbd325b1a046f40ffba75de/melodic/distribution.yaml#L4931-L4946

I haven't looked closely at whether melodic-devel has breaking changes, but a re-release on the melodic branch should be considered.

clalancette:
You are totally right; not sure how I messed that one up.  I think we'll need to actually look at cross-porting changes between kinetic-devel and melodic-devel to make sure we have everything.

clalancette:
So doing some research, it looks like https://github.com/ros/robot_state_publisher/pull/82 was the reason we wanted to release from a different branch.  Otherwise, the kinetic-devel branch and melodic-devel branch are broadly similar.

I'll point out that when we merged #82, there were only 2 PRs necessary to fix downstream breakage.  One of them was just lack of a header file, which was merged and which really wasn't directly involved.  The other was https://github.com/leggedrobotics/xpp/pull/5, which was closed because we never did break the API and thus it didn't apply.

So I see 3 different ways we can handle this from here:
1.  We can either knowingly break the API of robot_state_publisher by releasing from melodic-devel as-is with #82 applied.  There aren't a lot of downstream consumers that subclass it, so the breakage would probably be minimal (but not zero).
1.  Revert #82 completely from the melodic-devel branch, which will restore the API there and make it so that things worked as they once did.
1.  Revert the API change from #82, but don't do the tf_prefix resolving internally anymore.  We'll have a vestigial API parameter, but we can remove that in Noetic.

@sloretz Thoughts?

jschleicher:
I'd strongly encourage to get rid of the tf dependency in melodic (as it clutters minimal images with e.g. graphviz https://github.com/ros/geometry/issues/189).

So I prefer option 1.

jschleicher:
@sloretz @clalancette Any progress or decision?

clalancette:
All right, so we decided what to do here.

While I would love to release melodic-devel as-is into Melodic, there is too much risk of breaking existing Melodic users.  That's my fault, I made the mistake when we released the package.

[...]

PR #114, opened 2019-08-27
Noetic: Add joint_state_listener to the catkin package LIBRARIES (#112)

clalancette:
This is a forward-port of #112 to noetic-devel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ find_package(Eigen3 REQUIRED)
 find_package(urdfdom_headers REQUIRED)
 
 catkin_package(
-  LIBRARIES ${PROJECT_NAME}_solver
+  LIBRARIES ${PROJECT_NAME}_solver joint_state_listener
   INCLUDE_DIRS include
   DEPENDS roscpp rosconsole rostime tf2_ros tf2_kdl kdl_parser orocos_kdl urdfdom_headers
 )


sloretz:
@clalancette since there haven't been any noetic specific changes yet, I'd vote to fast-forward `noetic-devel` to `melodic-devel` 50f135253d92633ba7657f667c6ab37a6f83b731 instead

sloretz (commented):
@clalancette since there haven't been any noetic specific changes yet, I'd vote to fast-forward `noetic-devel` to `melodic-devel` 50f135253d92633ba7657f667c6ab37a6f83b731 instead

clalancette:
> @clalancette since there haven't been any noetic specific changes yet, I'd vote to fast-forward `noetic-devel` to `melodic-devel` [50f1352](https://github.com/ros/robot_state_publisher/commit/50f135253d92633ba7657f667c6ab37a6f83b731) instead

Sorry I didn't see this earlier.  There is one important difference between melodic-devel and noetic-devel, and that is commit 55203e5070c3be5ecb81a8d0d1d0ef8dcc5e54e6 .  If you remember, that was the commit that broke ABI and that we were *supposed* to release into Melodic, but which I messed up.  So Melodic actually ended up being released from the Kinetic branch originally.

As part of resolving #108, then, I reverted that commit from melodic-devel, then re-released into Melodic from melodic-devel.  The important bit is that noetic-devel does *not* have that commit reverted, so it breaks ABI.  As long as we remember to release from the noetic-devel branch properly, that change will be in Noetic.

All of that is context for the fact that I think we do need this port on the noetic-devel branch.

sloretz (approved):

clalancette:
CI fails because we don't yet have kdl_parser released in Noetic, but we know that.  So going ahead and merging, thanks for the review @sloretz .

PR #115, opened 2019-08-30
Added support for custom RobotStatePublisher implementations (API/ABI break)

peci1:
Retargeted #106 to noetic-devel.

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -48,7 +48,12 @@ using namespace KDL;
 using namespace robot_state_publisher;
 
 JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m, const urdf::Model& model)
-  : state_publisher_(createRobotStatePublisher(tree, model)), mimic_(m)
+  : JointStateListener(std::make_shared<RobotStatePublisher>(tree, model), m)
+{
+}
+
+JointStateListener::JointStateListener(const std::shared_ptr<RobotStatePublisher>& rsp, const MimicMap& m)
+  : state_publisher_(rsp), mimic_(m)
 {
   ros::NodeHandle n_tilde("~");
   ros::NodeHandle n;
@@ -147,13 +152,6 @@ void JointStateListener::callbackJointState(const JointStateConstPtr& state)
       last_publish_time_[state->name[i]] = state->header.stamp;
     }
   }
-}
-
-std::shared_ptr<RobotStatePublisher>
-JointStateListener::createRobotStatePublisher(const KDL::Tree& tree,
-                                              const urdf::Model& model)
-{
-  return std::make_shared<RobotStatePublisher>(tree, model);
 }
 
 // ----------------------------------
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -48,7 +48,7 @@ using namespace KDL;
 using namespace robot_state_publisher;
 
 JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m, const urdf::Model& model)
-  : state_publisher_(tree, model), mimic_(m)
+  : state_publisher_(createRobotStatePublisher(tree, model)), mimic_(m)
 {
   ros::NodeHandle n_tilde("~");
   ros::NodeHandle n;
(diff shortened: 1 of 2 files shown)


src/joint_state_listener.cpp
```diff
@@ -48,7 +48,7 @@ using namespace KDL;
 using namespace robot_state_publisher;
 
 JointStateListener::JointStateListener(const KDL::Tree& tree, const MimicMap& m, const urdf::Model& model)
-  : state_publisher_(tree, model), mimic_(m)
+  : state_publisher_(createRobotStatePublisher(tree, model)), mimic_(m)
```
clalancette: Looking at this, it actually seems a bit convoluted.  I get that you want to override the `RobotStatePublisher` class that is passed in, but having a virtual function that then gets called from inside the class seems too complicated.  Can we instead change the signature of `JointStateListener` to:

```
JointStateListener::JointStateListener(std::shared_ptr<RobotStatePublisher> rsp, const MimicMap& m) : state_publisher_(rsp), mimic_(m)
```

And then have callers do:

```
JointStateListener state_publisher(std::make_shared<RobotStatePublisher>(tree, model), mimic);
```

clalancette (changes requested):
I've got an idea to simplify this (since we are breaking ABI anyway, we may as well go ahead and break API).

On a totally separate note, I'm thinking of refactoring the ROS 2 port of `robot_state_publisher` to get rid of the split between the `RobotStatePublisher` and `JointStateListener` classes.  It seems overly complex to me, and it is preventing some further optimizations (like making the class composable).  Would that kind of refactor hurt your ability to customize it?  Do you have a pointer to how you are customizing it currently?

src/joint_state_listener.cpp
peci1: Great! If API breaks are allowed, I can imagine a much nicer code :) Go on for this idea.

My usage is in http://wiki.ros.org/dynamic_robot_state_publisher . If larger changes are allowed now, maybe we could think of integrating this functionality to the upstream. Basically it just exposes a service and dynamic reconfigure which allow to change/reload the robot model on the fly. But I'm aware ROS (1) is not prepared to work well with dynamically changing robot model... What about ROS 2? I think it'd be nice if all code from the beginning would account for the option that the model could change over time.

clalancette: > Great! If API breaks are allowed, I can imagine a much nicer code :) Go on for this idea.

API breaks are definitely allowed between releases.  In this case, I think we should do this API break for noetic since the resulting code will be cleaner.

> What about ROS 2? I think it'd be nice if all code from the beginning would account for the option that the model could change over time.

The current state of my rewrite of robot_state_publisher is in https://github.com/ros/robot_state_publisher/tree/ros2-refactor .  There, I've gotten rid of the `JointStateListener` class and just put all of the functionality into the `RobotStatePublisher` class.  I've also made it so that the user provides the contents of the URDF on the `robot_description` parameter, and publishes the results on topic by the same name.  In ROS 2, all parameters are dynamically reconfigurable, so the last remaining piece there is to reconfigure everything when the parameter changes (I just need to find the time to do that).

So in the end, the idea is that the user provides the URDF via a parameter, and that URDF can be changed at any time.  The rest of the network can then get notifications of changes to the URDF via the `robot_description` topic (or via `parameter_events`, but that has its own downsides).

What do you think about this setup?  Does that solve the problems you are hoping to solve with dynamic_robot_state_publisher?  If not, what else would you like to see it do?

peci1: The implementation looks good to me. To allow dynamic model changes, you'll definitely need to add a mutex and some complexity. Feel free to get inspired in my dynamic_robot_state_publisher.

One thing you'll have to think about is how to delete static TF frames. As we discussed in ros/geometry2#370, there is no good interpretation of a static TF frame disappearing. At least in ROS 1/tf2. I'm not sure if ROS2 hasn't added something like slow (but non-static) TFs which would get published on a latched topic...

peci1: And I think it'd be nice in the end to also do this change in ROS 1, probably for Noetic+.

sloretz: @clalancette Implemented Separate constructor to JointStateListener in 75f82a0

sloretz (approved):

clalancette (approved):

clalancette:
@sloretz Approved from me as well.  I'll let you merge it so that you can order things how you want with respect to your other PRs.

sloretz:
Thanks for the PR @peci1 !

PR #126, opened 2020-01-22
Refactor the ROS 2 code to be more modern

clalancette:
This is a major refactor of the robot_state_publisher code.  There are a few goals here:

1.  Finish up the port, including adding support for most of the parameters that were hard-coded before.
1.  Heavily cleanup the code to make it easier to understand.
1.  Switch to have robot_description provided as a parameter instead of a command-line argument.
1.  Allow the robot_description to be updated on the fly (via doing a "set" on the parameter).
1.  Allow robot_state_publisher to be composed.
1.  Add additional documentation on how to launch it.

With all of the above, this is a large API and ABI break, so is only appropriate for Foxy and forwards.  Once this is ready to go in, there will be another PR necessary to https://github.com/ros2/demos/blob/857c436e76bc6b40d32f096f6989bf51bb73f359/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup.launch.py#L28 to make it use the new parameter.

Reviews and thoughts on this welcome!


CMakeLists.txt
```diff
@@ -11,66 +11,52 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
-  src/joint_state_listener.cpp)
-target_link_libraries(
-  joint_state_listener
-  ${PROJECT_NAME}_solver
-  ${orocos_kdl_LIBRARIES})
+rclcpp_components_register_nodes(${PROJECT_NAME}
+  "robot_state_publisher::RobotStatePublisher")
 
 add_executable(
+  ${PROJECT_NAME}_node
```
mjcarroll: You may be able to avoid this boilerplate if you use `rclcpp_components_register_node` (https://github.com/ros2/rclcpp/blob/master/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L28), which will also create the executable for you.

include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -34,72 +34,85 @@
...
-#include "kdl/tree.hpp"
-
-#include "tf2_ros/static_transform_broadcaster.h"
-#include "tf2_ros/transform_broadcaster.h"
-#include "urdf/model.h"
-
-#include "std_msgs/msg/string.hpp"
-
-#include "rclcpp/rclcpp.hpp"
+typedef std::map<std::string, urdf::JointMimicSharedPtr> MimicMap;
```
mjcarroll: I generally prefer `using` to `typedef` in modern c++, but I don't think that it makes much of a difference here.

include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -34,72 +34,85 @@
...
-   */
-  RobotStatePublisher(
-    rclcpp::Node::SharedPtr node_handle, const KDL::Tree & tree,
-    const urdf::Model & model, const std::string & model_xml);
+  /// Constructor
+  explicit RobotStatePublisher(const rclcpp::NodeOptions & options);
 
   /// Destructor
-  virtual ~RobotStatePublisher() {}
+  ~RobotStatePublisher();
```
mjcarroll: Doesn't look like this does anything so ` = default` would be safe.

src/robot_state_publisher.cpp
```diff
@@ -68,110 +73,311 @@ geometry_msgs::msg::TransformStamped kdlToTransform(const KDL::Frame & k)
...
 {
-  // fprintf(stderr, "Publishing transforms for moving joints\n");
+  RCLCPP_DEBUG(get_logger(), "Publishing transforms for moving joints");
   std::vector<geometry_msgs::msg::TransformStamped> tf_transforms;
 
   // loop over all joints
-  for (auto jnt = joint_positions.begin(); jnt != joint_positions.end(); ++jnt) {
-    auto seg = segments_.find(jnt->first);
+  for (const std::pair<std::string, double> & jnt : joint_positions) {
+    std::map<std::string, SegmentPair>::iterator seg = segments_.find(jnt.first);
```
mjcarroll, line 240: Curious as to why you removed `auto` here

src/robot_state_publisher.cpp
```diff
@@ -68,110 +73,311 @@ geometry_msgs::msg::TransformStamped kdlToTransform(const KDL::Frame & k)
...
   std::vector<geometry_msgs::msg::TransformStamped> tf_transforms;
   geometry_msgs::msg::TransformStamped tf_transform;
 
   // loop over all fixed segments
-  for (auto seg = segments_fixed_.begin(); seg != segments_fixed_.end(); seg++) {
-    geometry_msgs::msg::TransformStamped tf_transform = kdlToTransform(seg->second.segment.pose(0));
-    rclcpp::Time now = clock_->now();
-    if (!use_tf_static) {
-      now = now + rclcpp::Duration(std::chrono::milliseconds(500));  // 0.5sec in NS
+  for (const std::pair<std::string, SegmentPair> & seg : segments_fixed_) {
```
mjcarroll, line 260: `auto`

mjcarroll (commented):
A few drive-by comments.

include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -34,72 +34,85 @@
...
   : segment(p_segment), root(p_root), tip(p_tip) {}
 
   KDL::Segment segment;
-  std::string root, tip;
+  std::string root;
+  std::string tip;
 };
 
-class RobotStatePublisher
+class RobotStatePublisher final : public rclcpp::Node
```
mjcarroll: I don't know if anybody was actually doing this before, but this removes the ability to subclass `RobotStatePublisher`, when previously that existed.

CMakeLists.txt
clalancette: Yeah, good point, I'll switch to that.

include/robot_state_publisher/robot_state_publisher.hpp
clalancette: This was a leftover from the old code.  But yeah, I'll change to `using`.

include/robot_state_publisher/robot_state_publisher.hpp
mjcarroll: A quick github search didn't reveal anything other than `robot_state_publisher`'s tests using that, so it may be safe: https://github.com/ros/robot_state_publisher/blob/melodic-devel/test/test_subclass.cpp#L60-L73

include/robot_state_publisher/robot_state_publisher.hpp
clalancette: Sure.

src/robot_state_publisher.cpp
clalancette: I find `auto` hard to read.  I never know what type the variable is, so I don't know what methods I can call on it.  I prefer to be explicit.  Same goes for below.

include/robot_state_publisher/robot_state_publisher.hpp
clalancette: Right.  I went back and forth on this a few times.  I believe some people *were* subclassing `RobotStatePublisher` in the past, but because of the total rearrangement of the class structure in this PR, that stuff would have to be totally rewritten now.

Really, I'm hoping that the reasons people were subclassing this could go away, and that stuff could be folded back in here.  But maybe that is optimistic and I shouldn't remove the possibility here.  I'm still on the fence.

CMakeLists.txt
```diff
@@ -80,25 +57,25 @@ install(DIRECTORY include/
...
 if(BUILD_TESTING)
   find_package(ament_cmake_cppcheck REQUIRED)
   find_package(ament_cmake_cpplint REQUIRED)
   find_package(ament_cmake_lint_cmake REQUIRED)
   find_package(ament_cmake_uncrustify REQUIRED)
-  # This forces cppcheck to consider all files in this project to be C++,
-  # including the headers which end with .h, which cppcheck would normally
-  # consider to be C instead.
-  ament_cppcheck(LANGUAGE "c++")
   ament_cpplint()
```
jacobperron: Consider replacing specific linter calls with `ament_lint_auto` or also add Python lint tests.

include/robot_state_publisher/robot_state_publisher.hpp
mjcarroll: I think that seems reasonable.

clalancette: I ended up making it sub-classable, and I removed the default destructor completely.

README.md
```diff
@@ -1,3 +1,21 @@
-# ROS2 Version of Robot State Publisher
+Robot State Publisher
+=====================
 
-This package contains a forked version of the original ROS Robot State Publisher with all modifications to compile within a ROS2 Ecosystem.
+This package contains the Robot State Publisher, a node and a class to publish the state of a robot to tf2.  Once the state gets published, it is available to all components in the system that also use tf2.  The package takes the joint angles of the robot as input and publishes the 3D poses of the robot links, using a kinematic tree model of the robot.
```
jacobperron: nit: one sentence per line

README.md
```diff
@@ -1,3 +1,21 @@
-# ROS2 Version of Robot State Publisher
+Robot State Publisher
+=====================
```
jacobperron, line 2: Looks like rst syntax in a markdown file. Consider renaming the file to have an `.rst` suffix.

README.md
```diff
@@ -1,3 +1,21 @@
...
 
-This package contains a forked version of the original ROS Robot State Publisher with all modifications to compile within a ROS2 Ecosystem.
+This package contains the Robot State Publisher, a node and a class to publish the state of a robot to tf2.  Once the state gets published, it is available to all components in the system that also use tf2.  The package takes the joint angles of the robot as input and publishes the 3D poses of the robot links, using a kinematic tree model of the robot.
+
+Examples showing how to pass the `robot_description` parameter using a launch file are available in the 'launch' subdirectory.
+
+Published Topics
+----------------
+* `robot_description` (`std_msgs/msg/String`) - The description of the robot URDF as a string.  Republishes the value set in the `robot_description` parameter, which is useful for getting informed of dynamic changes to the URDF.  Published using the "transient local" quality of service, so subscribers should also be "transient local".
```
jacobperron, line 12: I think it's also worth documenting it publishes to `/tf` and `/tf_static` in this list.

CMakeLists.txt
```diff
@@ -80,25 +57,51 @@ install(DIRECTORY include/
   DESTINATION include)
 
 install(
-  TARGETS ${PROJECT_NAME}
+  TARGETS ${PROJECT_NAME}_node
```
jacobperron: The node is already being installed by [rclcpp_components_register_node](https://github.com/ros2/rclcpp/blob/fdaf96f2171e70f3e013610aa44df2d7e9c866a3/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L53-L55). Although in retrospect maybe it shouldn't be, e.g. in case the user wants to build an executable only for testing :thinking:

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
+import os
```
jacobperron, line 33: Missing copyright/license

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+    return LaunchDescription([
+        node_robot_state_publisher,
+        test_exe_arg,
+        process_under_test,
+        launch_testing.util.KeepAliveProc(),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
+        ]), locals()
+
+
+class TestBufferClient(unittest.TestCase):
```
jacobperron: The name `TestBufferClient` looks vestigial of a copy from elsewhere?

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_termination(self, process_under_test, proc_info):
+        proc_info.assertWaitForShutdown(process=process_under_test, timeout=(10))
+
+
+@launch_testing.post_shutdown_test()
+class BufferClientTestAfterShutdown(unittest.TestCase):
```
jacobperron: Same here.

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+        parameters=[params]
+    )
+
+    return LaunchDescription([
+        node_robot_state_publisher,
+        test_exe_arg,
+        process_under_test,
+        launch_testing.util.KeepAliveProc(),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
+        ]), locals()
```
jacobperron:
```suggestion
    ]), locals()
```

I don't know if a linter will complain about this, but it's consistent with the style in the rest of this file.

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+        output='screen',
+        parameters=[params]
+    )
+
+    return LaunchDescription([
+        node_robot_state_publisher,
+        test_exe_arg,
+        process_under_test,
+        launch_testing.util.KeepAliveProc(),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
```
jacobperron: I think this is preferred now:
```suggestion
        launch_testing.actions.ReadyToTest(),
```

test/two_links_fixed_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+    return LaunchDescription([
+        node_robot_state_publisher,
+        test_exe_arg,
+        process_under_test,
+        launch_testing.util.KeepAliveProc(),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
+        ]), locals()
+
+
+class TestBufferClient(unittest.TestCase):
```
jacobperron: rename?

test/two_links_fixed_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_termination(self, process_under_test, proc_info):
+        proc_info.assertWaitForShutdown(process=process_under_test, timeout=(10))
+
+
+@launch_testing.post_shutdown_test()
+class BufferClientTestAfterShutdown(unittest.TestCase):
```
jacobperron: rename?

test/two_links_fixed_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+        output='screen',
+        parameters=[params]
+    )
+
+    return LaunchDescription([
+        node_robot_state_publisher,
+        test_exe_arg,
+        process_under_test,
+        launch_testing.util.KeepAliveProc(),
+        launch.actions.OpaqueFunction(function=lambda context: ready_fn()),
```
jacobperron:
```suggestion
        launch_testing.actions.ReadyToTest(),
```

test/two_links_moving_joint-launch.py
```diff
@@ -0,0 +1,56 @@
...
+        'test_exe',
+        description='Path to executable test',
+    )
+
+    process_under_test = launch.actions.ExecuteProcess(
+        cmd=[launch.substitutions.LaunchConfiguration('test_exe')],
+        output='screen',
+    )
+
+    urdf_file = os.path.join(os.path.dirname(__file__), 'two_links_moving_joint.urdf')
```
jacobperron, line 55: Correct me if I'm wrong, but it looks like this is the only difference with the other launch_testing file. Can we consolidate and make the urdf file name a launch argument?

test/test_two_links_moving_joint.cpp
```diff
@@ -0,0 +1,88 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2008, Willow Garage, Inc.
```
jacobperron: Is code moved/copied from somewhere or should this copyright line be updated?

src/robot_state_publisher.cpp
```diff
@@ -68,110 +73,308 @@ geometry_msgs::msg::TransformStamped kdlToTransform(const KDL::Frame & k)
...
+
+RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
+: rclcpp::Node("robot_state_publisher", options)
 {
-  // walk the tree and add segments to segments_
-  addChildren(tree.getRootSegment());
+  // get the XML
+  std::string urdf_xml = this->declare_parameter("robot_description", std::string(""));
+  if (urdf_xml.empty()) {
+    throw std::runtime_error("robot_description parameter must be provided");
```
jacobperron:
```suggestion
    throw std::runtime_error("robot_description parameter must not be empty");
```

src/robot_state_publisher.cpp
```diff
@@ -68,110 +73,308 @@ geometry_msgs::msg::TransformStamped kdlToTransform(const KDL::Frame & k)
...
+  }
+
+  // set publish frequency
+  double publish_freq = this->declare_parameter("publish_frequency", 20.0);
+  if (publish_freq > 1000.0) {
+    throw std::runtime_error("publish_frequency must be <= 1000.0");
```
jacobperron: Why is this a constraint? Seems arbitrary.

src/robot_state_publisher.cpp
```diff
@@ -68,110 +73,308 @@ geometry_msgs::msg::TransformStamped kdlToTransform(const KDL::Frame & k)
...
-  for (auto seg = segments_fixed_.begin(); seg != segments_fixed_.end(); seg++) {
-    geometry_msgs::msg::TransformStamped tf_transform = kdlToTransform(seg->second.segment.pose(0));
-    rclcpp::Time now = clock_->now();
-    if (!use_tf_static) {
-      now = now + rclcpp::Duration(std::chrono::milliseconds(500));  // 0.5sec in NS
+  for (const std::pair<std::string, SegmentPair> & seg : segments_fixed_) {
+    geometry_msgs::msg::TransformStamped tf_transform = kdlToTransform(seg.second.segment.pose(0));
+    rclcpp::Time now = this->now();
+    if (!use_tf_static_) {
+      now = now + rclcpp::Duration(std::chrono::milliseconds(500));
```
jacobperron, line 264: Why are we adding 500ms to the current time here?

jacobperron (commented):
Generally looks good! Some questions and suggestions below.

CMakeLists.txt
jacobperron: I don't think we necessarily need to remove this, but just pointing it out. Thoughts?

CMakeLists.txt
clalancette: Good idea.  Done in cf7fe91.

README.md
clalancette: Fixed in cf7fe91.

README.md
clalancette: This syntax does actually work in markdown, as far as I know.  You can see another example of how it renders [here](https://github.com/ros-drivers/phidgets_drivers/tree/dashing).

README.md
clalancette: Done in cf7fe91.

CMakeLists.txt
clalancette: You are totally correct, this is redundant.  Removed in cf7fe91.

test/two_links_moving_joint-launch.py
clalancette: Fixed in cf7fe91.

test/two_links_moving_joint-launch.py
clalancette: Yeah, good call.  Renamed in cf7fe91.

test/two_links_moving_joint-launch.py
clalancette: Renamed in cf7fe91.

test/two_links_moving_joint-launch.py
clalancette: Fixed in cf7fe91.

test/two_links_moving_joint-launch.py
clalancette: So I tried this, but it makes the tests fail for reasons I don't understand.  I'll have to come back to it later to investigate.

test/two_links_fixed_joint-launch.py
clalancette: Fixed in cf7fe91.

test/two_links_fixed_joint-launch.py
clalancette: Fixed in cf7fe91.

test/two_links_fixed_joint-launch.py
clalancette: Same as above; this didn't work for me.

test/two_links_moving_joint-launch.py
clalancette: It's not a bad idea.  I'll follow-up a little later on this.

test/test_two_links_moving_joint.cpp
clalancette: It's basically ported from the ROS 1 test in robot_state_publisher [here](https://github.com/ros/robot_state_publisher/blob/fe661c1503cad7fedc8c16be6edddb5eb18d5a65/test/test_two_links_moving_joint.cpp), so I thought the copyright should remain the same.

src/robot_state_publisher.cpp
clalancette: Fixed in cf7fe91.

src/robot_state_publisher.cpp
clalancette: It's because of the math below.  If the user specified 2000.0 for instance, they would end up dividing 1000.0 / 2000.0, which gets them 0.5 in double, but which ends up as 0 when cast to a uint64_t .  We could allow publishing at faster than 1000Hz by changing the type from `std::chrono::milliseconds` -> `std::chrono::microseconds`, but I didn't really see the point.

src/robot_state_publisher.cpp
clalancette: I honestly have no idea.  This is ported directly from https://github.com/ros/robot_state_publisher/blob/fe661c1503cad7fedc8c16be6edddb5eb18d5a65/src/robot_state_publisher.cpp#L117 ; I'll have to look into it later.

test/two_links_moving_joint-launch.py
clalancette: I figured it out.  There's another piece of magic I had to do to remove `ready_fn`.  Fixed in 9a6c78e47649fec084d66be1ac7caf3df39bd316.

test/two_links_fixed_joint-launch.py
clalancette: Fixed in 9a6c78e47649fec084d66be1ac7caf3df39bd316.

src/robot_state_publisher.cpp
jacobperron: Okay. It does seem unlikely someone will run into this constraint, and if they do we can always change to microseconds or nanoseconds.

We should probably guard against negative numbers too.

src/robot_state_publisher.cpp
jacobperron: :ok_hand:

src/robot_state_publisher.cpp
clalancette: Good point.  Fixed in db174336a159e586eb0dddbb2d6c3303fc610df4.

test/two_links_moving_joint-launch.py
clalancette: So I tried to do this, but I don't think it is possible (or at least, I don't know how to do it with the launch_testing API).  The problem is that inside the launch file, I need to do something like:

```
urdf_path = os.path.join(os.path.dirname(__file__), LaunchConfiguration('urdf_file').expand())
with open(urdf_path) as infp:
    robot_desc = infp.read()
params = {'robot_description': robot_desc}
```

Unfortunately, the `expand()` method on `LaunchConfiguration` just doesn't exist.  There is `perform()`, but that needs some sort of context that I don't know how to get.  Without that, I can't read in the robot_description that I need.  So I'm going to give up on this for now.

Any ideas on how to accomplish this are welcome!

src/robot_state_publisher.cpp
clalancette: I did a bit of archaeology in it, and found https://github.com/ros/robot_state_publisher/issues/46 .  In there, Jackie says that she considered this a hack, to be removed, but kept it for backwards-compatibility reasons.  It looks like she thought that we should just always default to use_tf_static, and remove the ability to shut it off.

That being said, I know we've had some complaints about this in the past, as splitting the tf tree between tf and tf_static can make it more difficult to get a full view of the tf tree.  So I'm on the fence about this.  My inclination is to leave it as-is, but I'm definitely open to hearing other opinions about it.

test/two_links_moving_joint-launch.py
jacobperron: I see the problem. Here's a [similar question](https://answers.ros.org/question/340705/access-launch-argument-in-launchfile-ros2/) with not quite the answer we're looking for. I feel like it should be possible; e.g. wrap the LaunchConfiguration substitution in another Substitution that does the file read. I don't have bandwidth atm to look into, so I'm happy to move forward with this PR as-is and if we figure it out we can revisit it.

src/robot_state_publisher.cpp
jacobperron: I tend to agree with removing the `use_tf_static` option and always publishing fixed frames as static (i.e. `/tf_static`), on the grounds that the split topics `/tf` and `/tf_static` are agreed on as a standard. IMO, tools that interact with the TF package (e.g. viewing the tree) should handle data coming from both channels.

Moving forward, I'd suggest deprecating the usage of `use_tf_static`; defaulting to `true` and warning the user if they set it to `false`.

My two cents.

jacobperron (approved):
LGTM with green CI

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9144/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4865/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9067/)

clalancette:
After doing the changes to dummy_robot_bringup in https://github.com/ros2/demos/pull/426 , there are 2 backwards compatibility concerns:

1.  The name of the standalone node switched from `robot_state_publisher` to `robot_state_publisher_node`.
1.  The URDF now has to be passed as a parameter, not a command-line argument, to the standalone node.

The first one is easy enough to fix, if we want to.  For the second one, we could make it backwards compatible (with a warning) if I hand-wrote the standalone node and did argument processing in there.  @jacobperron @sloretz thoughts on how important backwards compatibility is here?

jacobperron:
If there's not a strong argument for renaming the node, I'd vote for changing it back to `robot_state_publisher`. This should make it more straight forward for people not only coming from Eloquent, but also ROS 1.

Regarding the second point. You don't necessarily have to write the standalone node by hand. I ran into a similar situation when refactoring the image_tools demos and we added [this line](https://github.com/ros2/rclcpp/blob/fdaf96f2171e70f3e013610aa44df2d7e9c866a3/rclcpp_components/src/node_main.cpp.in#L32) so that non-ROS CLI args are passed to the node component (accessed with the NodeOptions `arguments()`, [for example](https://github.com/ros2/demos/blob/cfb9013c307a8ebe910d82ba76a46b5fe1ced0f1/image_tools/src/cam2image.cpp#L48)).

mjcarroll:
> Regarding the second point. You don't necessarily have to write the standalone node by hand. I ran into a similar situation when refactoring the image_tools demos and we added this line so that non-ROS CLI args are passed to the node component (accessed with the NodeOptions arguments(), for example).

Can we get this in the component tutorials/documentation somewhere? I think this is a pattern that people would like to know about.

jacobperron:
> Can we get this in the component tutorials/documentation somewhere? I think this is a pattern that people would like to know about.

Sure. At the time there was [discussion](https://github.com/ros2/rclcpp/pull/784#discussion_r320834394) as to whether we were actually abusing the NodeOptions API, which was originally intended for ROS-args. IMO, our usage is still a little awkward, since the rclcpp_components node is stripping out the ROS-args and only forwarding the non-ROS args. Maybe we should change the behaviour so that rclcpp_components doesn't strip away the ROS-args, or we can leave it as-is and document the special case for components. I can open a ticket to change the behaviour if I get a +1.

clalancette:
> If there's not a strong argument for renaming the node, I'd vote for changing it back to `robot_state_publisher`. This should make it more straight forward for people not only coming from Eloquent, but also ROS 1.

Yeah, agreed.  Will make that change.

> 
> Regarding the second point. You don't necessarily have to write the standalone node by hand. I ran into a similar situation when refactoring the image_tools demos and we added [this line](https://github.com/ros2/rclcpp/blob/fdaf96f2171e70f3e013610aa44df2d7e9c866a3/rclcpp_components/src/node_main.cpp.in#L32) so that non-ROS CLI args are passed to the node component (accessed with the NodeOptions `arguments()`, [for example](https://github.com/ros2/demos/blob/cfb9013c307a8ebe910d82ba76a46b5fe1ced0f1/image_tools/src/cam2image.cpp#L48)).

Oh, nice.  I'll play around with this, and make something that has backwards compatibility with the old stuff.

clalancette:
All right, backwards compatibility implemented.  I'm pretty close here, there is one last thing not working.  On Windows, the trick I'm using to pass the test executable name into the launch file doesn't work for some reason.  I'm still debugging that.

clalancette:
And tests fixed on Windows.  Here's another CI try:

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4867/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9070/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/9071/)

clalancette:
All right, all CI is green except for Windows Debug.  But the warning there already exists from the nightly: https://ci.ros2.org/view/nightly/job/nightly_win_deb/1509/

Unless there are further objections, I'm going to go ahead and merge this tomorrow morning.  Thanks again to @jacobperron and @mjcarroll for reviews!

CMakeLists.txt
```diff
@@ -76,6 +76,16 @@ if(BUILD_TESTING)
     sensor_msgs
     tf2_ros
   )
+  if(MSVC)
+    # CMake on Windows places executables in ${CMAKE_CURRENT_BINARY_DIR}/${RELEASE_TYPE},
```
jacobperron: I haven't noticed similar workarounds in other packages. What is different here?

jacobperron: Instead, I think we can use a generator expression like we do in [test_communication](https://github.com/ros2/system_tests/blob/7f018b29ebf3175bdbadcedfe4e45e25b8e08427/test_communication/CMakeLists.txt#L85). E.g.

```diff
- ${CMAKE_CURRENT_BINARY_DIR}/test_two_links_moving_joint
+ $<TARGET_FILE:test_two_links_moving_joint>
```

clalancette: Ah, that was the bit I didn't know.  OK, switched to TARGET_FILE, will run CI one more time.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9154/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4872/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7475/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9078/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/9079/)

jacobperron (approved):
LGTM!

Issue #125, opened 2020-01-22
tf_prefix no longer used in ROS2 node


pbeeson:
Is there a reason why tf_prefix can no longer be sent to the robot_state_publisher in ROS2?  

In ROS1, we use multiple robot_state_publishers as predictive displays for planning, each having a different namespace that maps /namespace/joint_states to the appropriate namespace in tf information.  So one urdf ends up creating multiple namespaced TF trees for the robot.

Attempting to migrate to ROS2 seemingly makes such functionality un-migratable.  
Is there an expected different way of doing this in ROS2?

sloretz:
`tf_prefix` was a feature of `tf` that was dropped in `tf2`. Only `tf2` has been ported to ROS 2, so there's no `tf_prefix` feature that `robot_state_publisher` can use.

One alternative is to remap `/tf` and `/tf_static` to `/namespace/tf` and `/namespace/tf_static`. Then have a node subscribe to those and publish transforms with prefixed frame names to the global `/tf` and `/tf_static` topics. This is similar to what ARIAC 2019 does: https://bitbucket.org/osrf/ariac/src/master/osrf_gear/src/tf2_relay.cpp

I'll close this in favor of ros/geometry2#32

pbeeson:
I really don't understand this.  If you look at the actual code for eloquent in robot_state_publisher, there is a tf_prefix variable that does get tacked onto both both the parent and child frame of the TFs being broadcast.  This does not used tf2 methods at all.  It's basically like a "namespace" that publishes on /tf without needing a messy TWO node solution like you suggest doing above (if only tf_preifix would read in as a parameter instead of being hard-coded to empty string).

That is, tf_prefix in robot_state_publisher DOES NOT depend on tf2 at all.  Just because it has the same name as an old method doesn't mean that the unrelated functionality needs to be removed.  RViz2 is still supporting tf_prefix in that when you enter a string there, it does exactly what I describe above and essentially prepends the string onto the frames.  So if RViz2 is supporting it, and the code is in the ROS2 port to prepend, and all you need is to read tf_prefix in as a parameter instead of hard-coding it to `""`, then I don't understand why it can't be there.  This seems 100% orthogonal to any tf2 deprecation.

sloretz:
>  If you look at the actual code for eloquent in robot_state_publisher, there is a tf_prefix variable that does get tacked onto both both the parent and child frame of the TFs being broadcast. This does not used tf2 methods at all.

[...]

PR #127, opened 2020-01-24
Add paramter back in for tf_prefix in robot_state_publisher

pbeeson:
As discussed in [issue 125](https://github.com/ros/robot_state_publisher/issues/125), the tf_prefix parameter for robot_state_publisher was a useful and reasonable parameter for the node.  The functionality still exists in the current ROS2 head except for the commented out parameter reading, which needed to be migrated to use ROS2 parameter syntax.  This PR adds the parameter back into ROS2, so that multiple robot_state_publishers can be loaded with the same URDF and push out info on the main /tf topic, but where each parent/child in "namespace" by the tf_prefix.

There is some confusion about tf_prefix, since tf_prefix() was removed from tf2, but this tf_prefix is largely orthogonal to that.  If desired, the parameter could be called tf_namespace, though pre-appending the "namespace" is still called tf_prefix in Rviz2 RobotModel plugin (and still works there the same as in RViz1).

### DIFF
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -77,10 +77,11 @@ JointStateListener::JointStateListener(
   n_tilde.param(tf_prefix_key, tf_prefix_, std::string(""));
   publish_interval_ = ros::Duration(1.0/max(publish_freq, 1.0));
   */
-
+  tf_prefix_="";
+  node->declare_parameter("tf_prefix");
+  node->get_parameter("tf_prefix",tf_prefix_);
   use_tf_static_ = true;
   ignore_timestamp_ = false;
-  tf_prefix_ = "";
   // auto publish_freq = 50.0;
   // publish_interval_ = std::chrono::seconds(1.0/std::max(publish_freq, 1.0));
   publish_interval_ = 1s;


raphaeldeimel:
The current state is utterly ridiculous when working with more than one robot sharing the same robot_description:

A) robot_state_publisher does not add prefixes anymore  - only alternative is to publish to separate topics
B) Rviz's RobotModel cannot listen to anything else than /tf.


In addition, all documentation still contains tf_prefix, and there is no mention of it being silently ignored in noetic. This will trip up a lot of melodic->noetic migrations.

raphaeldeimel:
To easee migration, you could add a compatibility node for prefixing&republishing transforms :

```
import rospy
import rosparam
from tf2_msgs.msg import TFMessage

rospy.init_node('tf_prefixer')

publisher = rospy.Publisher("/tf", TFMessage, queue_size=10)
prefix = rospy.get_param('tf_prefix')

def callback_tftopic(data):
    global publisher, prefix
    n_transforms = len(data.transforms)
    for i in range(n_transforms):
        data.transforms[i].header.frame_id = prefix + data.transforms[i].header.frame_id
        data.transforms[i].child_frame_id = prefix + data.transforms[i].child_frame_id
    publisher.publish(data)

listener = rospy.Subscriber("tf", TFMessage, callback_tftopic)

rospy.spin()
```
and mention it in the documentation

pbeeson:
Handled in #139, so closing here.

PR #134, opened 2020-04-07
Used keys for Orocos

ahcorde:
With this PR https://github.com/ros/rosdistro/pull/23841/files Orocos could be installed via rosdep

Signed-off-by: ahcorde <ahcorde@gmail.com>

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -28,7 +28,7 @@
 
   <build_depend>eigen</build_depend>
   <build_depend>kdl_parser</build_depend>
-  <build_depend version_gte="1.3.0">orocos_kdl</build_depend>
+  <build_depend>liborocos-kdl-dev</build_depend>
   <build_depend>rosconsole</build_depend>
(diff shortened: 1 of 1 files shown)


clalancette (commented):
I don't have a fundamental problem of doing this, but I am wondering how this is going to work in Noetic.

Is it going to be the case that we never release `orocos_kinematic_dynamics` into ROS Noetic?  If so, this will work fine.  If not, how is this going to work?  I just want to make sure I understand.  Thanks.

sloretz:
> Is it going to be the case that we never release orocos_kinematic_dynamics into ROS Noetic? If so, this will work fine. If not, how is this going to work? I just want to make sure I understand. Thanks.

Yup! Same for orocos-bfl. There's a note in the migration page here: http://wiki.ros.org/noetic/Migration#Use_Orocos_KDL_and_BFL_rosdep_keys

clalancette (approved):

PR #135, opened 2020-04-14
Add Missing rosbag test_depend

sloretz:
Used here: https://github.com/ros/robot_state_publisher/blob/806d7e0a46b2b22a3945209c4da167fe20df09ca/test/test_joint_states_bag.launch#L5

PR #136, opened 2020-04-15
Bump CMake version to avoid CMP0048

sloretz:
This bumps the minimum CMake version to 3.0.2, which is the [minimum supported by ROS Kinetic](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021) and new enough to default to the NEW behavior of [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html). This avoids a CMake warning when building and testing this package in Debian Buster and Ubuntu Focal.

ros/catkin#1052

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0.2)
 project(robot_state_publisher)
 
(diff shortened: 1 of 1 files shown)


sloretz:
Compiler warnings in CI are unrelated: https://github.com/ros/geometry2/pull/453

Test failure due to being unable to launch rosbag is fixed by #135

clalancette (approved):

PR #138, opened 2020-04-30
Claire/revert changes

claireyywang:
revert incorrect release changes for foxy on noetic branch


sloretz:
Ah oops, looks like an accidental release from the wrong branch. No harm to Noetic here. I think we can leave the changes and the 1.15.1 tag. The next Noetic release will just be one patch version higher than it would have otherwise.

claireyywang:
cool I'll close this then

PR #139, opened 2020-05-26
Restore tf_prefix support

rhaschke:
As discussed in #125, the `tf_prefix` parameter for `robot_state_publisher` was a useful and reasonable parameter. With the full migration from `tf` to `tf2` in #82, the tf_prefix functionality was removed as well, thus breaking behavior for many users, e.g. https://github.com/ros-visualization/rviz/issues/1507 and #125. 
A similar PR for ROS2 was filed in #127.
The TF and RobotModel displays in rviz still support the `tf_prefix` parameter as well (after intervention from some users).

This PR partially reverts #82.

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -82,7 +82,7 @@ void RobotStatePublisher::addChildren(const KDL::SegmentMap::const_iterator segm
   }
 }
 
-std::string stripSlash(const std::string & in)
+inline std::string stripSlash(const std::string & in)
 {
   if (in.size() && in[0] == '/')
   {
@@ -91,8 +91,21 @@ std::string stripSlash(const std::string & in)
   return in;
 }
 
+std::string resolve(const std::string &prefix, const std::string &frame)
+{
+  if (prefix.empty())
+    return stripSlash(frame);
+
+  std::string composite;
+  composite.reserve(prefix.length() + frame.length() + 1);
+  composite.append(stripSlash(prefix));
+  composite.append("/");
+  composite.append(stripSlash(frame));
+  return composite;
+}
+
 // publish moving transforms
-void RobotStatePublisher::publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time)
+void RobotStatePublisher::publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time, const std::string& tf_prefix)
 {
   ROS_DEBUG("Publishing transforms for moving joints");
   std::vector<geometry_msgs::TransformStamped> tf_transforms;
@@ -103,8 +116,8 @@ void RobotStatePublisher::publishTransforms(const std::map<std::string, double>&
     if (seg != segments_.end()) {
       geometry_msgs::TransformStamped tf_transform = tf2::kdlToTransform(seg->second.segment.pose(jnt->second));
       tf_transform.header.stamp = time;
-      tf_transform.header.frame_id = stripSlash(seg->second.root);
-      tf_transform.child_frame_id = stripSlash(seg->second.tip);
+      tf_transform.header.frame_id = resolve(tf_prefix, seg->second.root);
+      tf_transform.child_frame_id = resolve(tf_prefix, seg->second.tip);
       tf_transforms.push_back(tf_transform);
     }
     else {
@@ -115,7 +128,7 @@ void RobotStatePublisher::publishTransforms(const std::map<std::string, double>&
 }
 
 // publish fixed transforms
-void RobotStatePublisher::publishFixedTransforms(bool use_tf_static)
+void RobotStatePublisher::publishFixedTransforms(const std::string& tf_prefix, bool use_tf_static)
 {
   ROS_DEBUG("Publishing transforms for fixed joints");
   std::vector<geometry_msgs::TransformStamped> tf_transforms;
@@ -128,8 +141,8 @@ void RobotStatePublisher::publishFixedTransforms(bool use_tf_static)
     if (!use_tf_static) {
       tf_transform.header.stamp += ros::Duration(0.5);
     }
-    tf_transform.header.frame_id = stripSlash(seg->second.root);
-    tf_transform.child_frame_id = stripSlash(seg->second.tip);
+    tf_transform.header.frame_id = resolve(tf_prefix, seg->second.root);
+    tf_transform.child_frame_id = resolve(tf_prefix, seg->second.tip);
     tf_transforms.push_back(tf_transform);
   }
   if (use_tf_static) {
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -97,9 +97,8 @@ std::string resolve(const std::string &prefix, const std::string &frame)
     return stripSlash(frame);
 
   std::string composite;
-  composite.reserve(prefix.length() + frame.length() + 1);
+  composite.reserve(prefix.length() + frame.length());
   composite.append(stripSlash(prefix));
-  composite.append("/");
   composite.append(stripSlash(frame));
   return composite;
 }
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -75,6 +75,7 @@ protected:
   virtual void callbackJointState(const JointStateConstPtr& state);
   virtual void callbackFixedJoint(const ros::TimerEvent& e);
 
+  std::string tf_prefix_;
   ros::Duration publish_interval_;
   std::shared_ptr<RobotStatePublisher> state_publisher_;
   ros::Subscriber joint_state_sub_;
--- a/include/robot_state_publisher/robot_state_publisher.h
+++ b/include/robot_state_publisher/robot_state_publisher.h
@@ -80,8 +80,8 @@ public:
    * \param joint_positions A map of joint names and joint positions.
    * \param time The time at which the joint positions were recorded
    */
-  virtual void publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time);
-  virtual void publishFixedTransforms(bool use_tf_static = false);
+  virtual void publishTransforms(const std::map<std::string, double>& joint_positions, const ros::Time& time, const std::string& tf_prefix);
+  virtual void publishFixedTransforms(const std::string& tf_prefix, bool use_tf_static = false);
 
 protected:
   virtual void addChildren(const KDL::SegmentMap::const_iterator segment);
--- a/src/joint_state_listener.cpp
+++ b/src/joint_state_listener.cpp
@@ -71,6 +71,9 @@ JointStateListener::JointStateListener(const std::shared_ptr<RobotStatePublisher
   // ignore_timestamp_ == true, joins_states messages are accepted, no matter their timestamp
   n_tilde.param("ignore_timestamp", ignore_timestamp_, false);
   // get the tf_prefix parameter from the closest namespace
+  std::string tf_prefix_key;
+  n_tilde.searchParam("tf_prefix", tf_prefix_key);
+  n_tilde.param(tf_prefix_key, tf_prefix_, std::string(""));
   publish_interval_ = ros::Duration(1.0/std::max(publish_freq, 1.0));
 
   // Setting tcpNoNelay tells the subscriber to ask publishers that connect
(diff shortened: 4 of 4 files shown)


ana-GT:
Agreed on the need ot having the tf_prefix support restored for noetic.

ana-GT:
Is this PR going to be merged any time soon? This functionality is needed.

awinkler:
Yes, this was useful. I also have some repos (http://wiki.ros.org/xpp) where the noetic build fails due to this. It would be great to merge this and push to noetic.

UniBwTAS:
We also used this feature to visualize multiple cars in Rviz. We would be happy when this pull request gets merged.

schmittlema:
This would be very useful for [mushr](https://mushr.io/) to migrate to noetic! Please Merge

sloretz (changes requested):
It's clear that there is a large demand for this feature. After talking with some coworkers, it seems like the core part of it can be brought back as a `robot_state_publisher` feature. There's just a couple changes to make.

There was agreement about changing the parameter name to something other than `tf_prefix`. No matter what happens here, [`tf_prefix` is gone from upstream `tf2`](http://wiki.ros.org/tf2/Migration#Removal_of_support_for_tf_prefix). A name change communicates that behavior may be different. The name it self doesn't matter too much (`frame_prefix`, `frame_namespace`, etc) as long as it's different.

As a `robot_state_publisher` feature, I think the code that appends `/` between the prefix and the frame_id should be removed, and instead only the prefix should be concatenated with the frame id. `tf2` treats all frames as string literals, and `/` has no special meaning (except that it can't be the start of a frame_id). It wouldn't make sense to insert any extra characters into a `tf2` frame id, and someone wanting a `/` may still add a `/` to the end of the prefix.

I would recommend erroring if the prefix starts with `/` instead of stripping it.

rhaschke:
> I think the code that appends `/` between the prefix and the frame_id should be removed

Agreed, users can still add the slash (or any other separator) if they like.

> I would recommend erroring if the prefix starts with `/` instead of stripping it.

Not really agree. All other slashes are removed silently as well (e.g. from frame names).

> There was agreement about changing the parameter name to something other than `tf_prefix`.

I'm not sure about this. I agree that the generic support for tf_prefix was dropped, essentially because it wasn't consistently implemented by all applications. However, _robot_state_publisher_ and _rviz_ `have` implemented it correctly and the semantics stays exactly the same. So why changing the otherwise perfect name and force users to adapt?

sloretz:
>> I would recommend erroring if the prefix starts with / instead of stripping it.
>
> Not really agree. 

Oops, now I agree `robot_state_publisher` shouldn't error, at least in Noetic since it already strips the leading slash.

> All other slashes are removed silently as well (e.g. from frame names).

What silently removes all other slashes from frame names?

<details><summary>I don't see all other slashes being removed in `tf2` in ROS 1 or ROS 2</summary>

ROS 1 `static_transform_publisher` broadcasts slashes.
``` console
$ rosrun tf2_ros static_transform_publisher 0 0 0 0 0 0 1 /frame_starts_with_slash frame_has/slash_in_middle
[ INFO] [1603466592.355340614]: Spinning until killed publishing /frame_starts_with_slash to frame_has/slash_in_middle
```

Those slashes make it to the wire
```console
$ rostopic echo /tf_static
transforms: 
  - 
    header: 
      seq: 0
      stamp: 
        secs: 1603466592
        nsecs: 355257761
      frame_id: "/frame_starts_with_slash"
    child_frame_id: "frame_has/slash_in_middle"
    transform: 
      translation: 
        x: 0.0
        y: 0.0
        z: 0.0
      rotation: 
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
---
```

The `tf` tool `tf_echo` outputs frame name without the leading slash, but requires the leading slash present when asking for the frame name

```console
$ rosrun tf tf_echo /frame_starts_with_slash frame_has/slash_in_middle
At time 0.000
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
            in RPY (radian) [0.000, -0.000, 0.000]
            in RPY (degree) [0.000, -0.000, 0.000]
At time 0.000
$ rosrun tf tf_echo frame_starts_with_slash frame_has/slash_in_middle
Failure at 1603466659.664473823
Exception thrown:"fame_starts_with_slash" passed to lookupTransform argument target_frame does not exist. 
The current list of frames is:
Frame frame_has/slash_in_middle exists with parent frame_starts_with_slash.
```

Using ROS 2 the `static_transform_broadcaster` will accept frames with slashes and starting with a slash

``` console
$ ros2 run tf2_ros static_transform_publisher 0 0 0 0 0 0 1 /frame_starts_with_slash frame_has/slash_in_middle
[INFO] [1603466290.783300208] [static_transform_publisher_0Ff4MIb2j940JxY7]: Spinning until killed publishing transform from '/frame_starts_with_slash' to 'frame_has/slash_in_middle'
```

Those slashes also make it to the wire

```console
 ros2 topic echo --qos-durability transient_local --qos-depth 100 /tf_static
transforms:
- header:
    stamp:
      sec: 1603466290
      nanosec: 781788440
    frame_id: /frame_starts_with_slash
  child_frame_id: frame_has/slash_in_middle
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 0.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
---
```

But the `tf2` tool `tf2_echo` refuses to lookup a frame starting with a slash

```console
$ ros2 run tf2_ros tf2_echo /frame_starts_with_slash frame_has/slash_in_middle
[INFO] [1603466708.701367072] [tf2_echo]: Waiting for transform /frame_starts_with_slash ->  frame_has/slash_in_middle: Invalid frame ID "/frame_starts_with_slash" passed to canTransform argument target_frame - in tf2 frame_ids cannot start with a '/'
```

But it can be looked up if the leading slash is omitted.

```console
 ros2 run tf2_ros tf2_echo frame_starts_with_slash frame_has/slash_in_middle
[INFO] [1603467158.298894728] [tf2_echo]: Waiting for transform frame_starts_with_slash ->  frame_has/slash_in_middle: Invalid frame ID "frame_starts_with_slash" passed to canTransform argument target_frame - frame does not exist
At time 0.0
- Translation: [0.000, 0.000, 0.000]
- Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
```

</details>

None of the tools strip slashes from the middle. The `tf2` tools seem to allow leading slashes when broadcasting, but listening differs ROS 1 vs ROS 2. The ROS 1 tool requires the leading slash when listening, but the ROS 2 tool forbids it. Since `robot_state_publisher` is a broadcasting tool it would be consistent if it published the frame names unmodified, but I guess it should strip the leading slash of the "resolved" name in Noetic since the released version already does that.

> However, robot_state_publisher and rviz have implemented it correctly and the semantics stays exactly the same. So why changing the otherwise perfect name and force users to adapt?

I'm not 100% sure the semantics are exactly the same with the `tf` `tf_prefix` feature, but changing the name gives freedom to be different. With the switch to `tf2` I think it makes sense to try to use its conventions instead of `tf`'s.

rhaschke:
> All other slashes are removed silently as well (e.g. from frame names).

Of course, I meant all _leading_ slashes.

> I'm not 100% sure the semantics are exactly the same with the tf tf_prefix feature, but changing the name gives freedom to be different. With the switch to tf2 I think it makes sense to try to use its conventions instead of tf's.

I don't see how the semantics of `tf_prefix` would have changed? It's all about prefixing tf frames :wink:

sevangelatos:
We also use this feature in rviz to visualize multiple robots. We would be glad to see this merged!

sevangelatos:
@rhaschke Thank you for taking the time to restore `tf_prefix`. 

Initially, the PR retained the semantics of `tf_prefix` which was great for backwards compatibility.  But since you both agreed to  changing the handling of `/` characters, this results in a change of the behavior/semantics. I think the new handling of `/` makes better sense and it's OK to break backwards compatibility at the cost of some minor edits to our existing launch files.

Since it is needed to edit the launch files anyway, it is OK from my point of view to also change the parameter name. It makes the change of behavior easier to explain and document. And in practical terms, it also makes it easier to detect which files have been upgraded to the new behavior and which need to be upgraded.

Just my two cents, as someone who would love to see this merged. :-)

rhaschke:
I can understand your wish to merge this PR asap, but changing the name `tf_prefix` just to indicate a minor semantics change (removing a leading slash) is not justified imo. The other name suggestions are far worse in my view because they don't refer to TF anymore.

clalancette:
> I can understand your wish to merge this PR asap, but changing the name `tf_prefix` just to indicate a minor semantics change (removing a leading slash) is not justified imo. The other name suggestions are far worse in my view because they don't refer to TF anymore.

You are correct that the semantics of this are still "add a prefix to a tf name".  However, the main reason to move away from `tf_prefix` as a name is to make sure that this feature isn't confused with that earlier feature.

I don't have any particular name in mind for this feature here, as long as it is different than `tf_prefix`.  So I'm open to suggestions.

rhaschke:
> However, the main reason to move away from `tf_prefix` as a name is to make sure that this feature isn't confused with that earlier feature.

I don't see a (major) difference between these two features that justifies a name change.

Hugal31:
I agree with @rhaschke, but for the sake of making things move, I propose some alternative names:
- `prefix_tf`
- `prefix_frame_names_with`
- `prepend_frame_names_with`
- ~`tf_namespace`~
- `frame_names_prefix`
- `frames_in`
- `tf_in`

rhaschke:
+1 for prefix_tf: It's different, but very similar to the old tf_prefix

gavanderhoorn:
> +1 for prefix_tf: It's different, but very similar to the old tf_prefix

maybe even change to `prefix_tf_with`, makes for some nice XML:

```xml
<param name="prefix_tf_with" value="my_prefix/" />
```

---

Edit: oh, just saw that was one of the suggestions @Hugal31 already had in his list :)

ProgenitorX:
Wanted to express support to seeing this merged soon, regardless of what you want to call it. On board with `prefix_tf`.

rhaschke:
Ok, looks like there is some support for the name `prefix_tf_with`. I changed the name. @clalancette, could you have another look?

wxmerkt:
Friendly ping - what is the status of /decision on this PR?

oliwiermelon:
@sloretz, what is the status of this PR? My lab would also appreciate if this was resolved.

src/robot_state_publisher.cpp
```diff
@@ -91,8 +91,20 @@ std::string stripSlash(const std::string & in)
...
 
+std::string resolve(const std::string &prefix, const std::string &frame)
+{
+  if (prefix.empty())
+    return stripSlash(frame);
+
+  std::string composite;
+  composite.reserve(prefix.length() + frame.length());
+  composite.append(stripSlash(prefix));
+  composite.append(stripSlash(frame));
```
wxmerkt, line 102: I had to add `if (prefix.back() != '/') composite.append("/");` between L101 and L102: If we have tf_prefix "prefix" and frame "frame", with this version of "resolve", we'd have a frame published as "prefixframe" not "prefix/frame". Is this intended? If so, it would still be changed behaviour I believe?

I've tested this with a local version of @awinkler's xpp (where we are adding a workaround for this issue)

wxmerkt: I realise this relates to the discussion between @rhaschke and @sloretz above in the thread. As thus, 
(1) Is the consensus that all slashes - not only leading - should be stripped?
(2) Do we want to preserve the connecting-slash e.g. to work with the RobotModel plug-in in RViz?
(3) What happens if the prefix does not have a trailing slash (e.g. through prior user code specifying the prefix without a trailing slash)?

rakshith95:
Hello, is there any timeline on when this will be added?

SystemDiagnosticss:
Any updates in this case?

eaa3:
Hi there, are there any updates on bringing back this feature?

SystemDiagnosticss:
@rhaschke I am use your code with `prefix_tf_with` and namespace add to all robot links. But I cant add it to odom frame. Any idea?

tf_tree without namespace:
tf_tree

tf_tree with namespace:

tf_tree_2

Rayman:
> @rhaschke I am use your code with `prefix_tf_with` and namespace add to all robot links. But I cant add it to odom frame. Any idea?

The odom frame is not published by the robot state publisher. So you'd have to configure gazebo (so your urdf model) with the prefix.

i-am-neet:
Any updates? tf_prefix is very useful feature, just can't understand why removed it

i-am-neet:
> @rhaschke I am use your code with `prefix_tf_with` and namespace add to all robot links. But I cant add it to odom frame. Any idea?
> 
> tf_tree without namespace:
> tf_tree
> 
> tf_tree with namespace:
> 
> tf_tree_2

Hi,
I also tried some repos that fix the `tf_prefix` problem, but some plugins like `planar_move_plugin` doesn't work on those repos :((
So, I think the simple way is use `<xacro:arg>` for links in robot's xacro. FYR.

Any suggestion is welcome :D

sloretz:
I'll close this, as I think it's resolved by #169

PR #141, opened 2020-06-10
Make sure not to crash on an invalid URDF.

### LINKED ISSUE #140: Catch Exception when parsing invalid urdf
Hello, 
Imo it would be a great improvement to catch the thrown exception from `setupURDF(new_urdf)`. Atm the Ros node just dies if an invalid robot_description get loaded (on startup/or by an client). Especially if the robot_description is loaded via an client this is at least for me an unwanted behavior. 

In my fork I just added those lines, what fulfill my needs.

```
      std::string new_urdf = parameter.as_string();

      if (new_urdf.empty()) {
        result.successful = false;
        result.reason = "Empty URDF is not allowed";
        RCLCPP_INFO(get_logger(), "Empty URDF is not allowed");
        break;
      }

      try {
        setupURDF(new_urdf);
      }
      catch (std::runtime_error &exc) {
        RCLCPP_WARN(get_logger(), "Error parsing URDF");
        RCLCPP_WARN(get_logger(), exc.what());
        result.successful = false;
        result.reason = exc.what();
        break;
        }
      RCLCPP_INFO(get_logger(), "URDF OK");
```

regards mgangl

clalancette:
That is, once we are up and running, we only accept changes
to the URDF that successfully parse.  If the URDF doesn't parse,
return an error to the caller, print an error to the console,
and continue using the old one.

Also fix a small bug where we forgot to set the robot_description
parameter when we are using the fallback command-line strategy.

Fixes #140

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -98,6 +98,8 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
           in.seekg(0, std::ios::beg);
           in.read(&urdf_xml[0], urdf_xml.size());
           in.close();
+
+          this->set_parameter(rclcpp::Parameter("robot_description", urdf_xml));
         } else {
           throw std::system_error(
                   errno,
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6397/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9046/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11009/)

sloretz (approved):

PR #143, opened 2020-07-14
fix types in range loops to avoid copy due to different type

dirk-thomas:
Before: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/482/gcc/folder.1581101687/
After: [Build Status](https://ci.ros2.org/job/ci_linux/11464/)
* only unrelated warnings remaining

Regular CI builds testing the downstream packages (only with FastRTPS):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11465/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6671/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11359/)

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -175,14 +175,14 @@ void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
 
   // Initialize the mimic map
   mimic_.clear();
-  for (const std::pair<std::string, urdf::JointSharedPtr> & i : model_->joints_) {
+  for (const auto & i : model_->joints_) {
     if (i.second->mimic) {
       mimic_.insert(std::make_pair(i.first, i.second->mimic));
     }
   }
 
   KDL::SegmentMap segments_map = tree.getSegments();
-  for (const std::pair<std::string, KDL::TreeElement> & segment : segments_map) {
+  for (const auto & segment : segments_map) {
     RCLCPP_INFO(get_logger(), "got segment %s", segment.first.c_str());
   }
 
@@ -239,7 +239,7 @@ void RobotStatePublisher::publishTransforms(
   std::vector<geometry_msgs::msg::TransformStamped> tf_transforms;
 
   // loop over all joints
-  for (const std::pair<std::string, double> & jnt : joint_positions) {
+  for (const auto & jnt : joint_positions) {
     std::map<std::string, SegmentPair>::iterator seg = segments_.find(jnt.first);
     if (seg != segments_.end()) {
       geometry_msgs::msg::TransformStamped tf_transform =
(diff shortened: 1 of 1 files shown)


src/robot_state_publisher.cpp
```diff
@@ -175,14 +175,14 @@ void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
 
   // Initialize the mimic map
   mimic_.clear();
-  for (const std::pair<std::string, urdf::JointSharedPtr> & i : model_->joints_) {
```
sloretz, line 178: IIUC another way to fix would be to make the type match what the compiler expects. It looks like it expects the first element in the pair to be `const std::string`.

```
/home/jenkins-agent/workspace/nightly_linux_clang_libcxx/ws/src/ros/robot_state_publisher/src/robot_state_publisher.cpp:178:61: warning: loop variable 'i' has type 'const std::pair<std::string, urdf::JointSharedPtr> &' (aka 'const pair<basic_string<char, char_traits<char>, allocator<char> >, shared_ptr<urdf::Joint> > &') but is initialized with type 'std::__1::__map_iterator<std::__1::__tree_iterator<std::__1::__value_type<std::__1::basic_string<char>, std::__1::shared_ptr<urdf::Joint> >, std::__1::__tree_node<std::__1::__value_type<std::__1::basic_string<char>, std::__1::shared_ptr<urdf::Joint> >, void *> *, long> >::value_type' (aka 'pair<const std::__1::basic_string<char>, std::__1::shared_ptr<urdf::Joint> >') resulting in a copy [-Wrange-loop-construct]
```
If I'm reading the difference between the two types correctly

```c++
// Loop variable type
const pair<basic_string<char, char_traits<char>, allocator<char> >, shared_ptr<urdf::Joint> > &
// Actual type returned by the map iterator
pair<const std::__1::basic_string<char>, std::__1::shared_ptr<urdf::Joint> >
```

sloretz (commented):
The change LGTM from a functional perspective, but I'll let co-maintainer @clalancette chime in since I know auto has been intentionally avoided for clarity.

src/robot_state_publisher.cpp
dirk-thomas: I don't mind either way. 94b6f42a10094edb7827502230df314b98ae34c4 maintains the explicit type without using `auto`: [Build Status](https://ci.ros2.org/job/ci_linux/11469/) (only unrelated warnings remaining).

clalancette: Yeah, I like 94b6f42 better.  Looks good to me, thanks for the update.

clalancette (approved):

Karsten1987:
Can this be backported to Foxy?

dirk-thomas:
> Can this be backported to Foxy?

I don't see a reason why not.

Issue #147, opened 2020-09-10
ROS Noetic `tf_prefix` no longer supported


PedroACaldeira:
I've been trying to create a launch file for a multi robot system in noetic but I have issues with the robot_state_publisher package because I can't use the tf_prefix and as such, I have no way of organizing my robot system. I would like to know if there are any work around to make the launch file operable? For reference, I also made a question in the ROS Answers forum: https://answers.ros.org/question/360480/multi-robot-simulation-on-gazebo/

RaphvK:
See #139

oKermorgant:
Hi, 
I am facing the same issue while porting some tutorials / labs on ROS 2.

I hope the `tf_prefix` support pops up at some time. Meanwhile, a workaround that brings similar features is to use xacro prefixes for links. Basically all references to links in the URDF file are replaced by `"$(arg prefix)base_link"`.
The xacro file also takes an argument `<xacro:arg name="prefix" default=""/>`. 
You get the corresponding URDF by calling `xacro <file> prefix:=<prefix>/`.

If you are not the owner of the URDF / xacro files that you use, unfortunately it makes it mandatory to use a modified version. It would indeed be much simpler to have `robot_state_publisher` able to add a prefix on-the-fly without having to use a custom xacro file.

sloretz:
FYI some comments on bringing prefixing behavior back as a `robot_state_publisher` feature here: https://github.com/ros/robot_state_publisher/pull/139#pullrequestreview-506804970

SystemDiagnosticss:
Any updates in this case?

sloretz:
Resolved in #169

PR #159, opened 2021-03-04
Restore tf_prefix support for ROS2

smnogar:
Relevant issue: https://github.com/ros/robot_state_publisher/issues/147
Relevant PRs: Noetic support: https://github.com/ros/robot_state_publisher/pull/139 Closed ROS2 PR: https://github.com/ros/robot_state_publisher/pull/127

This is a quick and dirty implementation of tf_prefix for foxy. Once the string formatting issues are resolved in https://github.com/ros/robot_state_publisher/pull/139, I recommend adding them here. If this is too duplicative feel free to close this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,6 @@ if(BUILD_TESTING)
   add_executable(test_two_links_fixed_joint test/test_two_links_fixed_joint.cpp)
   ament_target_dependencies(test_two_links_fixed_joint
     rclcpp
-    sensor_msgs
     tf2_ros
   )
   target_link_libraries(test_two_links_fixed_joint ${GTEST_LIBRARIES})
@@ -83,7 +82,6 @@ if(BUILD_TESTING)
   add_executable(test_two_links_fixed_joint_prefix test/test_two_links_fixed_joint_prefix.cpp)
   ament_target_dependencies(test_two_links_fixed_joint_prefix
     rclcpp
-    sensor_msgs
     tf2_ros
   )
   target_link_libraries(test_two_links_fixed_joint_prefix ${GTEST_LIBRARIES})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,16 @@ if(BUILD_TESTING)
   add_launch_test(test/two_links_fixed_joint-launch.py
     ARGS "test_exe:=$<TARGET_FILE:test_two_links_fixed_joint>")
 
+  add_executable(test_two_links_fixed_joint_prefix test/test_two_links_fixed_joint_prefix.cpp)
+  ament_target_dependencies(test_two_links_fixed_joint_prefix
+    rclcpp
+    sensor_msgs
+    tf2_ros
+  )
+  target_link_libraries(test_two_links_fixed_joint_prefix ${GTEST_LIBRARIES})
+  add_launch_test(test/two_links_fixed_joint_prefix-launch.py
+    ARGS "test_exe:=$<TARGET_FILE:test_two_links_fixed_joint_prefix>")
+
   add_executable(test_two_links_moving_joint test/test_two_links_moving_joint.cpp)
   ament_target_dependencies(test_two_links_moving_joint
     rclcpp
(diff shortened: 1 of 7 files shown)


clalancette:
@smnogar First of all, I apologize for the very long delay in looking at this.

What I ended up doing here was make some changes to your PR, partially based off of the discussion that has been going on in #139.  In particular I did the following:

1. I rebased this on top of the `ros2` branch, which is where we land any new changes first.  Once we have it there, we can consider backporting it to Galactic and Foxy.
1. I renamed the prefix from `tf_prefix` to `frame_prefix`.  See https://github.com/ros/robot_state_publisher/pull/139#pullrequestreview-506804970 for why we think this should not just be `tf_prefix`.
1. I removed any special handling of `/` here.  Since this is now just a pure prefix, and the slash doesn't mean anything special to tf2 anymore, there is no reason to handle it specially here.
1. I added in a test for this to ensure we are getting what we expect.

With that, I'm pretty happy with this.  I'd appreciate a look from you to make sure this meets your needs.  @sloretz When you have some time, a review here would be appreciated as well.

test/test_two_links_fixed_joint_prefix.cpp
```diff
@@ -0,0 +1,77 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <gtest/gtest.h>
+
+#include <rclcpp/rclcpp.hpp>
+#include <sensor_msgs/msg/joint_state.hpp>
```
sloretz: nit, unused include?

CMakeLists.txt
```diff
@@ -80,6 +80,16 @@ if(BUILD_TESTING)
   add_launch_test(test/two_links_fixed_joint-launch.py
     ARGS "test_exe:=$<TARGET_FILE:test_two_links_fixed_joint>")
 
+  add_executable(test_two_links_fixed_joint_prefix test/test_two_links_fixed_joint_prefix.cpp)
+  ament_target_dependencies(test_two_links_fixed_joint_prefix
+    rclcpp
+    sensor_msgs
```
sloretz: nit, unnecessary dependency?

sloretz (approved):
A couple nitpicks, otherwise LGTM!

test/test_two_links_fixed_joint_prefix.cpp
clalancette: Yes, definitely.  Removed it from here and from `test_two_links_fixed_joint.cpp` in e83f247

CMakeLists.txt
clalancette: Yes, definitely.  Removed it from here and from `test_two_links_fixed_joint` in e83f247

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14684/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9436/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12349/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14827/)

clalancette:
All right, nits are fixed, CI is happy.  I'm going to go ahead and merge this one in.  @smnogar If you'd like to have this in Foxy or Galactic, please pipe up here and/or feel free to open backport PRs for those.  Thanks for the contribution.

padhupradheep:
Is it possible to have the frame_prefix support for foxy as well ? I guess this is a cool feature, that we would love to have!

clalancette:
> Is it possible to have the frame_prefix support for foxy as well ? I guess this is a cool feature, that we would love to have!

So looking back over this code, I realize that the big problem with backporting is that it is not ABI compatible.  That is, it changes the size of to `RobotStatePublisher` class, so we can't backport it as-is.

That said, we can probably do something slightly hacky where we don't declare the parameter, but instead allow undeclared parameters to be set and just get the parameter as needed.  That won't change the ABI.  Let me give that a shot and see if I can make it work.

PR #161, opened 2021-04-17
Stop rejecting unknown parameters.

clalancette:
This was never correct, but in Galactic this is fatal in several
ways.  Just remove it.

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -402,10 +402,6 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
         timer_ = this->create_wall_timer(
           publish_interval_ms_, std::bind(&RobotStatePublisher::publishFixedTransforms, this));
       }
-    } else {
-      result.successful = false;
-      result.reason = "Invalid parameter";
-      break;
     }
   }
 


src/robot_state_publisher.cpp
```diff
@@ -402,10 +402,6 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
         timer_ = this->create_wall_timer(
           publish_interval_ms_, std::bind(&RobotStatePublisher::publishFixedTransforms, this));
       }
-    } else {
-      result.successful = false;
-      result.reason = "Invalid parameter";
-      break;
```
sloretz, line 408: > this is fatal in several ways.

If I understand correctly, this block of code is unnecessary because `robot_state_publisher` declares its parameters and does not allow undeclared parameters. Why is it fatal though? Is it because of `use_sim_time`?

clalancette: > If I understand correctly, this block of code is unnecessary because `robot_state_publisher` declares its parameters and does not allow undeclared parameters. Why is it fatal though? Is it because of `use_sim_time`?

Actually, "fatal" isn't quite the right term.  I should not have used that.

The problem is that in Galactic, there are more "internal" parameters (specifically around qos_overrides), and so having this bit here means that you can never set any of those dynamically.  This is actually a problem with `use_sim_time` as well, but since that isn't something you usually set at runtime (only at startup time), we never ran into it before.

sloretz (approved):

PR #166, opened 2021-06-24
corrected publish_frequency default in README

schnili:
Corrected the default `publish_frequency` value in the README as it was missing its unit and was also not reflecting the correct value that is specified [in the code](https://github.com/ros/robot_state_publisher/blob/d8b471b742bd16734184167622eca3c6dce10980/src/robot_state_publisher.cpp#L119).


clalancette (approved):

sloretz (approved):

sloretz:
Merged without CI since it's a README change only

PR #169, opened 2021-07-01
Reimplement tf_prefix in Noetic exactly as it was in Melodic

sloretz:
This is an alternative to #139.  The goal of this PR is to make ROS Noetic `robot_state_publisher` behave the same as the ROS Melodic version.  It's different from #139 in these ways:

1. It calls the parameter `tf_prefix`
2. It reimplements the tf1 prefix and frame slash handling exactly as it behaves in ROS Melodic
2. It does not break ABI

While it doesn't break ABI, it effectively breaks using `JointStateListener` with a custom `RobotStatePublisher`. In this PR `JointStateListener` calls non-virtual versions of `publishFixedTransforms()` and `publishTransforms()`. Users who subclass `RobotStatePublisher`, override `publishFixedTransforms()` or `publishTransforms()`, and pass their subclass instance to a vanilla `JointStateListener` will observe their callbacks never being called.

**Note:** ROS 2 `robot_state_publisher` uses a different approach. It has a parameter `frame_prefix` that only appends a string to the frame name - no tf1 style special handling of slashes. See #159 for details.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,6 @@ if (CATKIN_ENABLE_TESTING)
   add_rostest_gtest(test_subclass ${CMAKE_CURRENT_SOURCE_DIR}/test/test_subclass.launch test/test_subclass.cpp)
   target_link_libraries(test_subclass ${catkin_LIBRARIES} ${PROJECT_NAME}_solver joint_state_listener)
 
-  install(FILES test/one_link.urdf test/pr2.urdf test/two_links_fixed_joint.urdf test/two_links_moving_joint.urdf DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/test)
   install(FILES test/one_link.urdf test/pr2.urdf test/two_links_fixed_joint.urdf test/two_links_moving_joint.urdf test/frames_and_slashes.urdf DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/test)
 
 endif()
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -84,6 +87,7 @@ if (CATKIN_ENABLE_TESTING)
   target_link_libraries(test_subclass ${catkin_LIBRARIES} ${PROJECT_NAME}_solver joint_state_listener)
 
   install(FILES test/one_link.urdf test/pr2.urdf test/two_links_fixed_joint.urdf test/two_links_moving_joint.urdf DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/test)
+  install(FILES test/one_link.urdf test/pr2.urdf test/two_links_fixed_joint.urdf test/two_links_moving_joint.urdf test/frames_and_slashes.urdf DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/test)
```
clalancette, line 81: This line is redundant, isn't it?  That is, shouldn't you just be able to add `test/frames_and_slashes.urdf` to the preceding line and be done with it?

clalancette (approved):
I've left one nit, but otherwise, this is a really good implementation.  One other issue is that CI seems to be failing.  I'm going to approve anyway, assuming that you'll fix both of these issues before merging.

jmirabel:
Any chances this will be merged soon ?

It has been more than a year since the issue has been raised and proposal have been made. I think this functionality is useful and many people would love to use this again.

CI on focal fails because it failed to open the [bag file](http://wiki.ros.org/robot_state_publisher/data?action=AttachFile&do=get&target=joint_states_indexed.bag). Does it ring a bell ?

mlanighan:
Any updates on a timeline for when this will be merged?

CMakeLists.txt
clalancette: This comment is still open.

sloretz: Rebased and removed redundant install in e817dcf

clalancette (approved):
Looks good to me assuming CI goes green.

PR #171, opened 2021-07-06
Galactic: Restore tf frame prefix support for ROS 2 (#159)

clalancette:
quick and dirty frame_prefix support for foxy

Also make it so we don't handle / specially.
Add in tests and documentation as well.

Co-authored-by: Chris Lalancette <clalancette@openrobotics.org>

This is an ABI compatible version of #159 for Galactic.

### DIFF
--- a/test/test_two_links_fixed_joint_prefix.cpp
+++ b/test/test_two_links_fixed_joint_prefix.cpp
@@ -39,22 +39,25 @@
 
 #define EPS 0.01
 
-TEST(test_publisher, test_two_joints)
+TEST(test_publisher, test_two_links_fixed_joint_prefix)
 {
-  auto node = rclcpp::Node::make_shared("rsp_test_two_links_fixed_joint");
+  auto node = rclcpp::Node::make_shared("rsp_test_two_links_fixed_joint_prefix");
 
   rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_SYSTEM_TIME);
   tf2_ros::Buffer buffer(clock);
   tf2_ros::TransformListener tfl(buffer, node, true);
 
-  for (unsigned int i = 0; i < 100 && !buffer.canTransform("link1", "link2", rclcpp::Time()); i++) {
+  for (unsigned int i = 0;
+    i < 100 && !buffer.canTransform("my_prefix/link1", "my_prefix/link2", rclcpp::Time()); i++)
+  {
     std::this_thread::sleep_for(std::chrono::milliseconds(100));
   }
 
-  ASSERT_TRUE(buffer.canTransform("link1", "link2", rclcpp::Time()));
+  ASSERT_TRUE(buffer.canTransform("my_prefix/link1", "my_prefix/link2", rclcpp::Time()));
   ASSERT_FALSE(buffer.canTransform("base_link", "wim_link", rclcpp::Time()));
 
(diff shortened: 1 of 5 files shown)


test/test_two_links_fixed_joint_prefix.cpp
```diff
@@ -0,0 +1,76 @@
...
+TEST(test_publisher, test_two_links_fixed_joint_prefix)
+{
+  auto node = rclcpp::Node::make_shared("rsp_test_two_links_fixed_joint_prefix");
+
+  rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_SYSTEM_TIME);
+  tf2_ros::Buffer buffer(clock);
+  tf2_ros::TransformListener tfl(buffer, node, true);
+
+  for (unsigned int i = 0;
+    i < 100 && !buffer.canTransform("my_prefix/link1", "my_prefix/link2", rclcpp::Time()); i++)
```
sloretz, line 51: nit, `++i`

test/test_two_links_fixed_joint_prefix.cpp
```diff
@@ -0,0 +1,76 @@
...
+  {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+
+  ASSERT_TRUE(buffer.canTransform("my_prefix/link1", "my_prefix/link2", rclcpp::Time()));
+  ASSERT_FALSE(buffer.canTransform("base_link", "wim_link", rclcpp::Time()));
```
sloretz, line 57: What's this assertion checking?

sloretz (commented):
LGTM with green CI and a couple comments resolved

test/test_two_links_fixed_joint_prefix.cpp
clalancette: Hm, I guess that a random transform doesn't work?  I'm honestly not sure, I copied that out of https://github.com/ros/robot_state_publisher/blob/ros2/test/test_two_links_fixed_joint.cpp , which is a copy/rewrite of https://github.com/ros/robot_state_publisher/blob/noetic-devel/test/test_two_links_fixed_joint.cpp#L90 .  Looking back at the history (both here in robot_state_publisher and in robot_model, where this forked out of), that particular check has been there since June 2010 :).

test/test_two_links_fixed_joint_prefix.cpp
clalancette: This is a copy of the code from https://github.com/ros/robot_state_publisher/blob/ros2/test/test_two_links_fixed_joint.cpp .  So if we were going to change this, I'd rather change it in all of the places on the `ros2` branch, and then possibly backport it to galactic.

sloretz: roger - sounds like a lot of work for very little benefit. I'll ignore it here.

sloretz (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14959/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9699/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12636/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15122/)

PR #172, opened 2021-07-20
Cleanup the documentation in the RobotStatePublisher class.

clalancette:
Make it so that it renders a lot more nicely.  We also rearrange
class methods here so that they are in the same order as in the
source file, but this should make no difference to API or ABI.

### DIFF
--- a/include/robot_state_publisher/robot_state_publisher.hpp
+++ b/include/robot_state_publisher/robot_state_publisher.hpp
@@ -75,7 +75,8 @@ public:
   explicit RobotStatePublisher(const rclcpp::NodeOptions & options);
 
 protected:
-  /** Setup the URDF for use.
+  /// Setup the URDF for use.
+  /**
    * This method first parses the URDF into an internal representation.
    * Based on that representation, it then generates the list of joint segments
    * and mimic pairs that it needs during runtime.  Finally, it publishes
@@ -85,13 +86,15 @@ protected:
    */
   void setupURDF(const std::string & urdf_xml);
 
-  /** Recursive method to add all children to the internal segment list.
+  /// Recursive method to add all children to the internal segment list.
+  /**
    *
    * \param[in] segment An iterator to the SegmentMap to add to the internal segment list.
    */
   void addChildren(const KDL::SegmentMap::const_iterator segment);
 
-  /** Publish transforms to /tf2.
+  /// Publish transforms to /tf2.
+  /**
    * This method is called by callbackJointState() when new transforms are available and need to be published.
    *
    * \param[in] joint_positions A map of joint names to joint positions.
@@ -101,11 +104,11 @@ protected:
     const std::map<std::string, double> & joint_positions,
     const builtin_interfaces::msg::Time & time);
 
-  /** Publish fixed transforms at startup time to /tf2_static.
-   */
+  /// Publish fixed transforms at startup time to /tf2_static.
   void publishFixedTransforms();
 
-  /** The callback that is called when a new JointState message is received.
+  /// The callback that is called when a new JointState message is received.
+  /**
    * This method examines the incoming JointStates and applies a series of checks to
    * see if new transforms should be published.  If so, it calls publishTransforms() to do so.
    *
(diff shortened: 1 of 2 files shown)


include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -51,56 +51,117 @@ using MimicMap = std::map<std::string, urdf::JointMimicSharedPtr>;
...
+  /** The callback that is called when a new JointState message is received.
+   * This method examines the incoming JointStates and applies a series of checks to
+   * see if new transforms should be published.  If so, it calls publishTransforms() to do so.
+   *
+   * \param[in] state The JointState message that was delivered.
+   */
   void callbackJointState(const sensor_msgs::msg::JointState::SharedPtr state);
-  void setupURDF(const std::string & urdf_xml);
+
+  /** The callback that is called when parameter on the node are changed.
```
sloretz:
```suggestion
  /** The callback that is called when parameters on the node are changed.
```

include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -51,56 +51,117 @@ using MimicMap = std::map<std::string, urdf::JointMimicSharedPtr>;
...
+   * This allows the class to dynamically react to changes in parameters.
+   *
+   * \param[in] parameters The list of parameters that are going to change.
+   */
   rcl_interfaces::msg::SetParametersResult parameterUpdate(
     const std::vector<rclcpp::Parameter> & parameters);
 
   std::map<std::string, SegmentPair> segments_;
+  ///< A map of dynamic segment names to SegmentPair structures
```
sloretz: Subjectively, I find it easier to read documentation before the member than after, but I wouldn't block on that.

include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -51,56 +51,117 @@ using MimicMap = std::map<std::string, urdf::JointMimicSharedPtr>;
...
-   * \param joint_positions A map of joint names and joint positions.
-   * \param time The time at which the joint positions were recorded
+  /** Setup the URDF for use.
+   * This method first parses the URDF into an internal representation.
+   * Based on that representation, it then generates the list of joint segments
+   * and mimic pairs that it needs during runtime.  Finally, it publishes
+   * the text of the URDF to the network on the /robot_description topic.
+   *
+   * \param[in] urdf_xml The string representing the URDF XML.
+   */
```
wjwwood: I don't know what the prevailing style in this repository is, and if this is your preference that's fine too, but the style in `rclcpp` (and other core packages) would be this:

```suggestion
  /// Setup the URDF for use.
  /**
   * This method first parses the URDF into an internal representation.
   * Based on that representation, it then generates the list of joint segments
   * and mimic pairs that it needs during runtime.  Finally, it publishes
   * the text of the URDF to the network on the /robot_description topic.
   *
   * \param[in] urdf_xml The string representing the URDF XML.
   */
```

include/robot_state_publisher/robot_state_publisher.hpp
wjwwood: Same.

wjwwood (approved):
lgtm, two non-blocking comments

include/robot_state_publisher/robot_state_publisher.hpp
clalancette: I also agree.  The problem is that rosdoc2 didn't like that.  If I put the comment before the variable, then they would all be "shifted" down by 1 (e.g. it would say "A map of dynamic segment names to SegmentPair structures" for the `segments_fixed_` variable).

Let me poke at it again briefly and see if I can figure out what is going on there.

clalancette: Ah, it turns out I didn't understand Doxygen syntax.

```
///< document the member that comes *before* the comment
```

```
/// document the member that comes *after* the comment
```

I switched to the latter now, and it all seems happier.

wjwwood (approved):

clalancette:
Given that this changes documentation only, I'm going to merge just based on the Rpr job.  Thanks for the reviews!

PR #177, opened 2021-10-07
Change joint_states subscription to SensorDataQoS

adeschamps:
This changes the subscription from reliable to best_effort. This is helpful when visualizing a robot that's running on a wifi-connected device, and eliminates a warning when starting up drivers on a robot that publishes with best_effort.

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -144,7 +144,7 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
 
   // subscribe to joint state
   joint_state_sub_ = this->create_subscription<sensor_msgs::msg::JointState>(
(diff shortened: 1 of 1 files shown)


sloretz:
Newer versions of ROS 2 have an opt-in configurable QoS feature. Would you be willing to implement that here instead of changing the hard-coded QoS?

Design doc http://design.ros2.org/articles/qos_configurability.html
Example PR with the required changes: https://github.com/cra-ros-pkg/robot_localization/pull/657

sloretz:
Also, please target the `ros2` branch for new development. If it works in ROS Rolling, then it can be considered for backporting into a stable distro like ROS Galactic.

adeschamps:
Closing in favour of https://github.com/ros/robot_state_publisher/pull/179.

Thanks for the link to that design doc and the example PR -- both were very helpful.

PR #179, opened 2021-10-08
Make joint_states subscription QoS configurable; default to SensorDataQoS

adeschamps:
This changes the subscription default from `reliable` to
`best_effort`. This is helpful when visualizing a robot that's running
on a wifi-connected device, and eliminates a warning when starting up
drivers on a robot that publishes with `best_effort`.

The subscription defaults to `best_effort`, but can be overridden with
the following config:

```yaml
robot_state_publisher:
  ros__parameters:
    qos_overrides:
      /joint_states:
        subscription:
          reliability: reliable
```

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -136,11 +136,16 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
 
   setupURDF(urdf_xml);
 
+  auto subscriber_options = rclcpp::SubscriptionOptions();
+  subscriber_options.qos_overriding_options =
+    rclcpp::QosOverridingOptions::with_default_policies();
+
   // subscribe to joint state
   joint_state_sub_ = this->create_subscription<sensor_msgs::msg::JointState>(
-    "joint_states", 10, std::bind(
-      &RobotStatePublisher::callbackJointState, this,
-      std::placeholders::_1));
+    "joint_states",
+    rclcpp::SensorDataQoS(),
+    std::bind(&RobotStatePublisher::callbackJointState, this, std::placeholders::_1),
+    subscriber_options);
 
   publishFixedTransforms();
 


src/robot_state_publisher.cpp
```diff
@@ -136,11 +136,16 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
...
+  subscriber_options.qos_overriding_options =
+    rclcpp::QosOverridingOptions::with_default_policies();
+
   // subscribe to joint state
   joint_state_sub_ = this->create_subscription<sensor_msgs::msg::JointState>(
-    "joint_states", 10, std::bind(
-      &RobotStatePublisher::callbackJointState, this,
-      std::placeholders::_1));
+    "joint_states",
+    rclcpp::SensorDataQoS(),
```
clalancette, line 146: I think I'd rather keep this as `reliable` by default, but you now have the option to switch it to `best-effort` if you want.  At least, that's the most compatible with the current setup and least surprising.

clalancette (changes requested):

src/robot_state_publisher.cpp
adeschamps: I think `best-effort` makes sense as a default, and I wouldn't consider it a breaking change -- according to the table in the [QoS compatibilities](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html#qos-compatibilities) section, a `best-effort` subscriber can still receive messages from a `reliable` publisher. Changing the default here lays the groundwork for allowing publishers to change their default in a non-breaking way in a future version.

clalancette: > according to the table in the [QoS compatibilities](https://docs.ros.org/en/galactic/Concepts/About-Quality-of-Service-Settings.html#qos-compatibilities) section, a `best-effort` subscriber can still receive messages from a `reliable` publisher.

While that is true, best-effort doesn't intuitively seem like the correct default for the `joint_state`.  If you miss a `joint_state` update, then you won't update the tf2 tree appropriately, and there is nothing that guarantees that the publisher will republish that joint in the near future.  That is, this doesn't seem to act like a Sensor, where you know you will get another update about the state of the world soon.

adeschamps: The systems I've worked with publish `joint_state` at regular intervals, such as the [create_driver](https://github.com/AutonomyLab/create_robot/blob/f0a8dea560f83d148223f62bf8b0734775e09be3/create_driver/src/create_driver.cpp#L286) for my hobby Roomba and the platforms I've used at work.

That said, it sounds like you have some counterexamples in mind, where changing this default would cause issues. Even though I still think that `best-effort` would be a more suitable default in most situations, I'm happy to switch the default back in order to avoid such issues.

sloretz: If the QoS is configurable, I think the default QoS values becomes less important. Those creating a system where the QoS matters can set it to what they need. Either value seems fine to me.

I only recall working with systems that publish joint states regularly, but I could imagine a node that didn't (maybe a node reading wheel encoder values only updates when the encoder value changes).

The compatibility table of a best effort subscription seems like a nice default user experience to me. It can match with reliable and best-effort publishers without additional configuration. Whether or not that's a good idea for the specific system is something I would expect the person integrating the software to be responsible for.

clalancette: > I only recall working with systems that publish joint states regularly, but I could imagine a node that didn't (maybe a node reading wheel encoder values only updates when the encoder value changes).

Yeah, this is the counter-example that I had in mind, and the one that I've run into.

That said, in the particular robot where I saw this, it being best-effort probably would not have mattered.  The node publishing `joint_state` and `robot_publisher` were both on the same machine, so there is little likelihood that DDS would have dropped the data anyway.

OK, 2-1 against me, I'm OK with changing the default.  I'm running out of time to do it today, but I'll review again and run CI tomorrow.

clalancette (approved):
Looks good to me.  I'll run CI next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10178/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15702/)

PR #180, opened 2021-10-22
Change how parameter updates are handled.

clalancette:
set_parameter callbacks should not have any side-effects.
This is most easy to understand if you think about the fact
that those callbacks can be chained; if callback A has a side-effect,
but then callback B rejects the entire set_parameter call, then
the nodes will be in an inconsistent state.

Instead, change it so that the set_parameter callback checks that
the incoming arguments are valid, and does nothing else.  We
setup a callback to listen to the /parameter_events topic;
at that point, we know that the update has been accepted by
all set_parameter callbacks, so it is safe for side-effects to happen.

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -162,35 +162,30 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
     std::bind(&RobotStatePublisher::onParameterEvent, this, std::placeholders::_1));
 }
 
-std::pair<std::shared_ptr<urdf::Model>, KDL::Tree> RobotStatePublisher::parseURDF(
-  const std::string & urdf_xml)
+KDL::Tree RobotStatePublisher::parseURDF(const std::string & urdf_xml, urdf::Model & model)
 {
-  auto model = std::make_shared<urdf::Model>();
-
   // Initialize the model
-  if (!model->initString(urdf_xml)) {
+  if (!model.initString(urdf_xml)) {
     throw std::runtime_error("Unable to initialize urdf::model from robot description");
   }
 
   // Initialize the KDL tree
   KDL::Tree tree;
-  if (!kdl_parser::treeFromUrdfModel(*model, tree)) {
+  if (!kdl_parser::treeFromUrdfModel(model, tree)) {
     throw std::runtime_error("Failed to extract kdl tree from robot description");
   }
 
-  return std::make_pair(model, tree);
+  return tree;
 }
 
 void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
 {
-  std::pair<std::shared_ptr<urdf::Model>, KDL::Tree> modeltree = parseURDF(urdf_xml);
-
-  std::shared_ptr<urdf::Model> model = std::get<0>(std::move(modeltree));
-  KDL::Tree tree = std::get<1>(modeltree);
+  urdf::Model model;
+  KDL::Tree tree = parseURDF(urdf_xml, model);
 
   // Initialize the mimic map
   mimic_.clear();
-  for (const std::pair<const std::string, urdf::JointSharedPtr> & i : model->joints_) {
+  for (const std::pair<const std::string, urdf::JointSharedPtr> & i : model.joints_) {
     if (i.second->mimic) {
       mimic_.insert(std::make_pair(i.first, i.second->mimic));
     }
@@ -215,7 +210,7 @@ void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
 
 // add children to correct maps
 void RobotStatePublisher::addChildren(
-  std::shared_ptr<urdf::Model> model,
+  const urdf::Model & model,
   const KDL::SegmentMap::const_iterator segment)
 {
   const std::string & root = GetTreeElementSegment(segment->second).getName();
@@ -225,8 +220,8 @@ void RobotStatePublisher::addChildren(
     const KDL::Segment & child = GetTreeElementSegment(children[i]->second);
     SegmentPair s(GetTreeElementSegment(children[i]->second), root, child.getName());
     if (child.getJoint().getType() == KDL::Joint::None) {
-      if (model->getJoint(child.getJoint().getName()) &&
-        model->getJoint(child.getJoint().getName())->type == urdf::Joint::FLOATING)
+      if (model.getJoint(child.getJoint().getName()) &&
+        model.getJoint(child.getJoint().getName())->type == urdf::Joint::FLOATING)
       {
         RCLCPP_INFO(
           get_logger(), "Floating joint. Not adding segment from %s to %s.",
@@ -370,7 +365,8 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
 
       // And that we can successfully parse it
       try {
-        parseURDF(new_urdf);
+        urdf::Model dummy_model;
+        parseURDF(new_urdf, dummy_model);
       } catch (const std::runtime_error & err) {
         RCLCPP_WARN(get_logger(), "%s", err.what());
         result.successful = false;
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -44,7 +44,6 @@
 #include <urdf/model.h>
 
 #include <chrono>
-#include <cstring>
 #include <fstream>
 #include <functional>
 #include <map>
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -122,10 +122,10 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
   }
 
   // set frame_prefix
-  frame_prefix_ = this->declare_parameter("frame_prefix", "");
+  this->declare_parameter("frame_prefix", "");
 
   // ignore_timestamp_ == true, joint_state messages are accepted, no matter their timestamp
-  ignore_timestamp_ = this->declare_parameter("ignore_timestamp", false);
+  this->declare_parameter("ignore_timestamp", false);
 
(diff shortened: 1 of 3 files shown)


src/robot_state_publisher.cpp
```diff
@@ -144,30 +144,48 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
 
   publishFixedTransforms();
 
+
```
hidmic: @clalancette nit: unrelated?

src/robot_state_publisher.cpp
```diff
@@ -144,30 +144,48 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
...
+    this->get_node_topics_interface(),
+    std::bind(&RobotStatePublisher::onParameterEvent, this, std::placeholders::_1));
 }
 
-void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
+std::pair<std::shared_ptr<urdf::Model>, KDL::Tree> RobotStatePublisher::parseURDF(
+  const std::string & urdf_xml)
 {
-  model_ = std::make_unique<urdf::Model>();
+  auto model = std::make_shared<urdf::Model>();
```
hidmic: @clalancette why using an `std::shared_ptr`? Also, why is it no longer necessary to keep `model` alive?

src/robot_state_publisher.cpp
```diff
@@ -144,30 +144,48 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
...
 
+  return std::make_pair(model, tree);
+}
+
+void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
+{
+  std::pair<std::shared_ptr<urdf::Model>, KDL::Tree> modeltree = parseURDF(urdf_xml);
+
+  std::shared_ptr<urdf::Model> model = std::get<0>(std::move(modeltree));
+  KDL::Tree tree = std::get<1>(modeltree);
```
hidmic: @clalancette meta, fyi: if we bump this to C++17, we can use [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) instead.

src/robot_state_publisher.cpp
```diff
@@ -191,7 +209,9 @@ void RobotStatePublisher::setupURDF(const std::string & urdf_xml)
 }
 
 // add children to correct maps
-void RobotStatePublisher::addChildren(const KDL::SegmentMap::const_iterator segment)
+void RobotStatePublisher::addChildren(
+  std::shared_ptr<urdf::Model> model,
```
hidmic: @clalancette does model ownership need to be transferred to `addChildren`? If not, consider using a raw pointer or a reference.

src/robot_state_publisher.cpp
```diff
@@ -332,64 +355,58 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
 
   for (const rclcpp::Parameter & parameter : parameters) {
     if (parameter.get_name() == "robot_description") {
-      // First make sure that it is still a string
-      if (parameter.get_type() != rclcpp::ParameterType::PARAMETER_STRING) {
-        result.successful = false;
-        result.reason = "URDF must be a string";
-        break;
-      }
```
hidmic, line 346: @clalancette if this is gone, what's type checking parameters now?

src/robot_state_publisher.cpp
```diff
@@ -332,64 +355,58 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
...
-        result.reason = "frame_prefix must be a string";
-        break;
-      }
-      frame_prefix_ = parameter.as_string();
-    } else if (parameter.get_name() == "ignore_timestamp") {
-      if (parameter.get_type() != rclcpp::ParameterType::PARAMETER_BOOL) {
-        result.successful = false;
-        result.reason = "ignore_timestamp must be a boolean";
-        break;
-      }
```
hidmic, line 378: @clalancette I see we're disallowing `ignore_timestamp` and `frame_prefix` configuration. Why?

src/robot_state_publisher.cpp
clalancette: Ah, this is an interesting bit of code.  I'll answer the questions in reverse order.

As for why we don't need the model alive, it's because after the parsing and the setup of the URDF, we no longer need to keep it around.  That is, the rest of the class never accesses that particular member variable.

As for why using a `shared_ptr`, I struggled a bit with the pointer semantics here.  My original try at this was a `std::unique_ptr`, but that doesn't work because `addChildren` is recursive; we can't `std::move` the pointer into each of the children, because when we pop back up to the parent it would now be NULL (and further children wouldn't work anymore).

Another way to do this would be to make `model` be a stack variable, and then pass the raw pointer around (which I think you allude to below).  I didn't love this option, as it is only obvious that the raw pointer is "safe" if you examine all of the code.  But I guess there is a small enough amount of code here that it would be OK.

I don't mind hugely which way we go here.

src/robot_state_publisher.cpp
clalancette: Hm.  I've had enough trouble trying to enable C++17 with Windows that I don't think I'd choose this option right now.

src/robot_state_publisher.cpp
clalancette: Unrelated, I'll remove.

src/robot_state_publisher.cpp
clalancette: See my [comment above](https://github.com/ros/robot_state_publisher/pull/180/files#r734663723) on why this was done this way.  We can change it.

src/robot_state_publisher.cpp
clalancette: The internals of rclcpp now do type checking by default (unless you set `dynamic_typing = true`, which this class does not).  So these checks are now redundant.

src/robot_state_publisher.cpp
clalancette: It's not that we are disallowing them; it's that we don't need to check them anymore.  Since the `rclcpp` core is now doing the type checking for us, we no longer need to do it ourselves, and thus there are no situations in which we would reject an update that the core hasn't already rejected.

hidmic: Ok with the type checks, but we are removing [`ignore_timestamp_`](https://github.com/ros/robot_state_publisher/pull/180/files#diff-56a859e14fb1c861d179f60566024cf100f2a34b9c5d90fe1d727a04de5dd9f5L379) and [`frame_prefix_`](https://github.com/ros/robot_state_publisher/pull/180/files#diff-56a859e14fb1c861d179f60566024cf100f2a34b9c5d90fe1d727a04de5dd9f5L372) updates as well.

src/robot_state_publisher.cpp
hidmic: > I didn't love this option, as it is only obvious that the raw pointer is "safe" if you examine all of the code.

A reference would do as well.

hidmic (commented):
@clalancette thanks for the bump, completely forgot about this one. Left a couple questions.

src/robot_state_publisher.cpp
clalancette: Ah, of course!  I don't know why I didn't think of that, thanks.  I'll switch to a reference, that is far more reasonable.

src/robot_state_publisher.cpp
clalancette: Actually, given our discussion above this problem goes away.  So I'll mark this as "resolved" once I push that change.

src/robot_state_publisher.cpp
clalancette: This will also be resolved by using a reference.

src/robot_state_publisher.cpp
clalancette: Fixed in a0de001cb0a770f4da9b0c3aa1db471f4112c5b2

src/robot_state_publisher.cpp
clalancette: Removed in a0de001cb0a770f4da9b0c3aa1db471f4112c5b2

src/robot_state_publisher.cpp
clalancette: Removed in a0de001cb0a770f4da9b0c3aa1db471f4112c5b2

src/robot_state_publisher.cpp
clalancette: I double checked on this, and it is indeed the case that `rclcpp` is doing type-checking for us (at least on Rolling).  I tried to do:

```
ros2 param set /robot_state_publisher robot_description 5
```

And the result was:

```
Setting parameter failed: Wrong parameter type, parameter {robot_description} is of type {string}, setting it to {integer} is not allowed.
```

So we don't need these explicit checks anymore.  Marking this as resolved.

src/robot_state_publisher.cpp
clalancette: Oh, right, that's a really good point.  Thanks for pointing it out.

In point of fact, we don't really need to store `ignore_timestamp_` and `frame_prefix_` inside of the class; it is already being stored in the parameter database, so we can always just fetch it from there.  I'll go ahead and implement it that way instead so that we don't have to have the member variables.

clalancette: Fixed in adac2b7

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15836/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10513/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13501/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16121/)

hidmic (approved):

clalancette:
Thanks!  I'm going to stick with the non-const ref for now, we can always decide to change it later.  I'm going to go ahead and merge this one.

PR #182, opened 2021-11-12
Publish new fixed transforms when URDF is updated

### LINKED ISSUE #181: Static fixed transforms are not updated with robot_description parameter
In Galactic and later, when the robot URDF is updated at runtime by setting the `robot_description` parameter, static fixed transforms are currently not re-published to match the new model.

A workaround in Galactic is to use `tf_static := false`, but this obviously isn't ideal as the parameter has been deprecated, and it also has the side-effect of publishing fixed transforms 500ms into the future, still resulting in half a second of broken transform data after the URDF update.

A simple fix would be to call `publishFixedTransforms()` just after the URDF is updated in `parameterUpdate()` [here](https://github.com/ros/robot_state_publisher/blob/8b3f6b8e03432651e2b788128e17603f0972c677/src/robot_state_publisher.cpp#L357), to keep the transforms in line with the new URDF. The same fix should also work on Galactic, but should probably be predicated on `if (use_tf_static_)` to match the same behaviour in the initialiser.

I'm happy to put this into a PR, if this behaviour is indeed a bug.

RussellJoyce:
A new URDF may come with a new set of fixed transforms, so if the URDF
is updated through the `robot_description` parameter, we also need to
publish its new transforms.

Fixes #181.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,13 +68,12 @@ if(BUILD_TESTING)
   find_package(launch_testing_ament_cmake REQUIRED)
   ament_find_gtest()
 
-  include_directories(${GTEST_INCLUDE_DIRS})
-
   add_executable(test_two_links_fixed_joint test/test_two_links_fixed_joint.cpp)
   ament_target_dependencies(test_two_links_fixed_joint
     rclcpp
     tf2_ros
   )
+  target_include_directories(test_two_links_fixed_joint PRIVATE ${GTEST_INCLUDE_DIRS})
   target_link_libraries(test_two_links_fixed_joint ${GTEST_LIBRARIES})
   add_launch_test(test/two_links_fixed_joint-launch.py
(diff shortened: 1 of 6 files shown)


clalancette (approved):
@RussellJoyce Excellent, thanks for the fix!  I also added a test here to make sure that this functionality works (which it does with your fix in place).  This looks good to me, so I'm gonig to run CI on it.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15605/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10287/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13274/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15822/)

clalancette:
@Mergifyio backport galactic

PR #186, opened 2022-01-07
Fix include order for cpplint

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

### DIFF
--- a/include/robot_state_publisher/robot_state_publisher.hpp
+++ b/include/robot_state_publisher/robot_state_publisher.hpp
@@ -36,13 +36,13 @@
 #include <string>
 #include <vector>
 
-#include <builtin_interfaces/msg/time.hpp>
-#include <kdl/tree.hpp>
-#include <rcl_interfaces/msg/parameter_event.hpp>
-#include <rcl_interfaces/msg/set_parameters_result.hpp>
-#include <rclcpp/rclcpp.hpp>
-#include <sensor_msgs/msg/joint_state.hpp>
-#include <std_msgs/msg/string.hpp>
+#include "builtin_interfaces/msg/time.hpp"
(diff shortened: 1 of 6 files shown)


include/robot_state_publisher/robot_state_publisher.hpp
```diff
@@ -31,21 +31,21 @@
...
+#include <string>
+#include <vector>
+
 #include <builtin_interfaces/msg/time.hpp>
 #include <kdl/tree.hpp>
 #include <rcl_interfaces/msg/parameter_event.hpp>
 #include <rcl_interfaces/msg/set_parameters_result.hpp>
 #include <rclcpp/rclcpp.hpp>
 #include <sensor_msgs/msg/joint_state.hpp>
 #include <std_msgs/msg/string.hpp>
```
clalancette: While we are in here, we may as well change all of these to double-quotes, right?

src/robot_state_publisher.cpp
```diff
@@ -41,17 +51,7 @@
 #include <rclcpp_components/register_node_macro.hpp>
 #include <sensor_msgs/msg/joint_state.hpp>
 #include <std_msgs/msg/string.hpp>
```
clalancette: Same comment here.

include/robot_state_publisher/robot_state_publisher.hpp
jacobperron: Done b4d6088

src/robot_state_publisher.cpp
jacobperron: b4d6088

clalancette (approved):
Looks good with green CI.

PR #189, opened 2022-01-18
[Galactic backport]: Publish new fixed transforms when URDF is updated

### LINKED ISSUE #181: Static fixed transforms are not updated with robot_description parameter

### LINKED ISSUE #188: Static fixed transforms are not updated with robot_description parameter (Galactic Branch)
This error occurs on the galactic branch, it has been fixed on the ROS2 branch with [this issue](https://github.com/ros/robot_state_publisher/issues/181). But it has yet to been fixed on the Galactic branch.

Is it planned to update the Galactic branch soon with this fix ?

clalancette:
A new URDF may come with a new set of fixed transforms, so if the URDF
is updated through the `robot_description` parameter, we also need to
publish its new transforms.

Fixes #181

Signed-off-by: Russell Joyce <russell@russellj.co.uk>

This backports just the fix in #182 to Galactic; the rest of the cleanup and the tests I didn't backport (since it didn't seem necessary here).

Fixes #188

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -372,6 +372,7 @@ rcl_interfaces::msg::SetParametersResult RobotStatePublisher::parameterUpdate(
 
       try {
         setupURDF(new_urdf);
+        publishFixedTransforms();
       } catch (const std::runtime_error & err) {
         RCLCPP_WARN(get_logger(), "%s", err.what());
         result.successful = false;


BenjaminHug8:
Hi,

I just noticed than this PR is solving an issue I have and I was wondering when it will be merged?

Thank you!

mjcarroll (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16177/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10804/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16519/)

BenjaminHug8:
Thank you!

PR #197, opened 2022-05-09
Use the new clearTransforms() API in StaticTransformBroadcaster.

clalancette:
This will allow us to remove static joints at runtime.  While
we are in here, notice that we always call publishFixedTransforms()
in combination with setupURDF(), so just call one from the
other.

Depends on https://github.com/ros2/geometry2/pull/523 .

Still a draft because it needs further testing.


clalancette:
(the Rpr job is expected to fail until https://github.com/ros2/geometry2/pull/523 is merged and released)

Issue #200, opened 2022-10-08
Continuous requests to parameter server for tf_prefix with every joint_state message


nlamprian:
The robot_state_publisher continuously hits the parameter server with 3 requests for every joint_state message and in each fixed joint timer interval.

```
$ sudo tcpflow -c port 11311 -i lo

127.000.000.001.49382-127.000.001.001.11311: POST / HTTP/1.1
User-Agent: XMLRPC++ 0.7
Host: simtoru-0003:11311
Content-Type: text/xml
Content-length: 236

<?xml version="1.0"?>
<methodCall><methodName>searchParam</methodName>
<params><param><value>/robot_state_publisher</value></param><param><value>tf_prefix</value></param></params></methodCall>

127.000.001.001.11311-127.000.000.001.49382: HTTP/1.1 200 OK
Server: BaseHTTP/0.6 Python/3.8.10
Date: Sat, 08 Oct 2022 22:43:39 GMT
Content-type: text/xml
Content-length: 384

<?xml version='1.0'?>
<methodResponse>
<params>
<param>
<value><array><data>
<value><int>1</int></value>
<value><string>Found [/robot_state_publisher/tf_prefix]</string></value>
<value><string>/robot_state_publisher/tf_prefix</string></value>
</data></array></value>
</param>
</params>
</methodResponse>

127.000.000.001.49382-127.000.001.001.11311: POST / HTTP/1.1
User-Agent: XMLRPC++ 0.7
Host: simtoru-0003:11311
Content-Type: text/xml
Content-length: 296

<?xml version="1.0"?>
<methodCall><methodName>hasParam</methodName>
<params><param><value>/robot_state_publisher</value></param><param><value>/robot_state_publisher/tf_prefix</value></param></params></methodCall>

127.000.001.001.11311-127.000.000.001.49382: HTTP/1.1 200 OK
Server: BaseHTTP/0.6 Python/3.8.10
Date: Sat, 08 Oct 2022 22:43:39 GMT
Content-type: text/xml
Content-length: 307

<?xml version='1.0'?>
<methodResponse>
<params>
<param>
<value><array><data>
<value><int>1</int></value>
<value><string>/robot_state_publisher/tf_prefix</string></value>
<value><boolean>1</boolean></value>
</data></array></value>
</param>
</params>
</methodResponse>

127.000.000.001.49382-127.000.001.001.11311: POST / HTTP/1.1
User-Agent: XMLRPC++ 0.7
[...]

PR #201, opened 2022-10-12
cache tf_prefix

peroxyacyl:
On every joint_state callback, the node searches `tf_prefix` parameter, and it loads an unignorable impact on roscore processes.

This patch caches the result of the first search of tf_prefix. A downside is that we cannot update tf_prefix dynamically anymore (anyone does?).

ref. #200

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -87,7 +87,8 @@ protected:
   MimicMap mimic_;
   bool use_tf_static_;
   bool ignore_timestamp_;
-
+  std::string tf_prefix_;
+  bool tf_prefix_cached_;
 };
 }
 
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -87,8 +87,7 @@ protected:
   MimicMap mimic_;
   bool use_tf_static_;
   bool ignore_timestamp_;
-  std::string tf_prefix_;
-  bool tf_prefix_cached_;
+
 };
 }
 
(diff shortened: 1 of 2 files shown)


include/robot_state_publisher/joint_state_listener.h
```diff
@@ -87,7 +87,8 @@ class JointStateListener {
   MimicMap mimic_;
   bool use_tf_static_;
   bool ignore_timestamp_;
-
+  std::string tf_prefix_;
+  bool tf_prefix_cached_;
```
sloretz: Adding these members changes the size of the class, and that breaks ABI. This will have to be stored somewhere else so we can release it into Noetic.

One solution I've seen is to create a static global map in the `.cpp` file where the key is the `this` pointer, and the value is the things we want to store per class instance.

peroxyacyl: Thank you for your quick review!
I've put a minimal thread-safe map and stored `(this, tf_prefix)` in the callback. The entry is deleted in the destructor.

jobafr:
we're hitting this as well, would be great if it could get merged

sloretz:
Thank you for the PR!

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160). Every change comes with a risk of introducing regressions, and there isn't much time left to fix them. To make sure this PR doesn't introduce any regressions please:

* Describe how you tested this change
* Recruit at least one more person to review this PR and try it out on their system

----

@jobafr , since you're running into this, would you be willing to try it on your system and report if this fixes the issue for you?

clalancette:
Noetic is now EOL, so closing this one out.

PR #202, opened 2022-11-08
fixes the following bug: /tf topics timestamps switch from simulated time to system time when bagfile playbag is stopped and use_sim_time is set

C-NR:
while the playback of a bagfile is stopped and use_sime_time is true the /tf topics timestamps switch from simulated time to system time causing eventually lots of error messages (rviz2 and so on)


SGSSGene:
This is meant as a polite ping!
Is there any chance on progress?

PR #204, opened 2023-01-30
Update robot_state_publisher to C++17.

clalancette:
The main reason to do this is so that we can compile robot_state_publisher with the clang static analyzer.  As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version.  So set this as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(robot_state_publisher)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")


sloretz:
> Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version. So set this as well.

Interesting. Just curious, is `CMAKE_CXX_STANDARD_REQUIRED` still necessary if  `target_compile_features(tgt PUBLIC cxx_std_17)` is used instead of `CMAKE_CXX_STANDARD` when using clang?

clalancette:
> Interesting. Just curious, is `CMAKE_CXX_STANDARD_REQUIRED` still necessary if `target_compile_features(tgt PUBLIC cxx_std_17)` is used instead of `CMAKE_CXX_STANDARD` when using clang?

In short, no.  It seems like clang still defaults to using C++14 in that case, for reasons I don't quite understand.  This causes i to have lots of warnings for things like missing `std::optional` and `std::variant`.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18850/)

mjcarroll (approved):

PR #208, opened 2023-10-06
upgrade to C++14 for compatibility with Eigen 3.40.90

daizhirui:
With Eigen 3.40.90 installed from source, we need to change the C++ standard from 11 to 17. Otherwise, we get error: `#error This compiler appears to be too old to be supported by Eigen`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 3.0.2)
 project(robot_state_publisher)
 
 include(CheckCXXCompilerFlag)
-unset(COMPILER_SUPPORTS_CXX11 CACHE)
+unset(COMPILER_SUPPORTS_CXX17 CACHE)
 if(MSVC)
   # https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version
   # MSVC will fail the following check since it does not have the c++11 switch
   # however, c++11 is always enabled (the newer /std:c++14 is enabled by default)
-  check_cxx_compiler_flag(/std:c++11 COMPILER_SUPPORTS_CXX11)
-  if(COMPILER_SUPPORTS_CXX11)
-    add_compile_options(/std:c++11)
+  # Oct. 06 2023: upgrade to c++17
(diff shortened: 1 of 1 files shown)


daizhirui:
The CI seems to be broken: `KeyError: 'ros-noetic-catkin'`.

CMakeLists.txt
```diff
@@ -2,22 +2,23 @@ cmake_minimum_required(VERSION 3.0.2)
...
   # https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version
   # MSVC will fail the following check since it does not have the c++11 switch
   # however, c++11 is always enabled (the newer /std:c++14 is enabled by default)
-  check_cxx_compiler_flag(/std:c++11 COMPILER_SUPPORTS_CXX11)
-  if(COMPILER_SUPPORTS_CXX11)
-    add_compile_options(/std:c++11)
+  # Oct. 06 2023: upgrade to c++17
+  check_cxx_compiler_flag(/std:c++17 COMPILER_SUPPORTS_CXX17)
+  if(COMPILER_SUPPORTS_CXX17)
+    add_compile_options(/std:c++17)
```
sloretz: Ah, I'd recommend replacing this logic with `target_compile_features(target_name PUBLIC cxx_std_14)`. Here's an example from ROS 2: https://github.com/ros2/rclcpp/blob/fcbe64cff4bea3109531254ceb2955dc4b1bb320/rclcpp/CMakeLists.txt#L196

sloretz (changes requested):
Thanks for the PR!

[Noetic targets C++ 14](https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025) , and ROS Noetic [is the last supported ROS 1 release](https://www.openrobotics.org/blog/2020/5/23/noetic-ninjemys-the-last-official-ros-1-release).

I can keep this open if you'd be willing to make it c++14 instead, but otherwise I'll close it.

CMakeLists.txt
daizhirui: I change it as you suggested.

daizhirui:
Changed to C++14 and tested with Eigen 3.4.90

sloretz (approved):
Thanks for iterating!

PR #210, opened 2023-12-07
Fix destructors should be public virtual

Rayman:
I encountered some compilation errors with clang. It complains that the destructor is not virtual. According to the cpp core guidelines a destructor should be either public and virtual or protected and non virtual. See also this link:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual

### DIFF
--- a/include/robot_state_publisher/joint_state_listener.h
+++ b/include/robot_state_publisher/joint_state_listener.h
@@ -69,7 +69,7 @@ public:
 
 
   /// Destructor
-  ~JointStateListener();
+  virtual ~JointStateListener();
 
 private:
   std::string getTFPrefix();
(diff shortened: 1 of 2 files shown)


include/robot_state_publisher/joint_state_listener.h
```diff
@@ -69,7 +69,7 @@ class JointStateListener {
 
 
   /// Destructor
-  ~JointStateListener();
+  virtual ~JointStateListener();
```
sloretz, line 72: This makes sense, but I think making it virtual will break ABI. We try to avoid breaking ABI in released distros. Would you be willing to make the destructor protected instead?

sloretz (changes requested):
Thank you for the PR!

sloretz:
Thank you for the PR!

This change breaks ABI, which we cannot do in a released ROS distro.

Given that [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), I'm going to close this.

PR #212, opened 2023-12-25
Fix reload after a description with a mimic joint

### LINKED ISSUE #211: [bug] Crash on refresh of robot_description parameter in case there is mimic joint in the original description
On iron (probably in rolling too):

When setting the `robot_description` after `robot_state_publisher` has been already started with a first description that contains a mimic joint, it crashes. More precisely, the following line: https://github.com/ros/robot_state_publisher/blob/f52c7ca4c85f0ac7cd11f6d371b64f48da78b562/src/robot_state_publisher.cpp#L186

## Minimal steps to reproduce
Start the following launch file:
```
import launch
import launch_ros.actions
from launch.actions import ExecuteProcess, LogInfo, RegisterEventHandler
from launch.event_handlers import OnProcessStart
from launch.substitutions import FindExecutable


def generate_launch_description():
    robot_desc = """\
<?xml version="1.0"?>
<robot name="test">
  <link name="base_link">
  </link>

  <link name="second_link">
  </link>

  <link name="third_link">
  </link>

  <joint name="base_to_second_joint" type="fixed">
    <origin rpy="1 0 0" xyz="0 1 0"/>
    <parent link="base_link"/>
    <child link="second_link"/>
  </joint>

  <joint name="base_to_third_joint" type="fixed">
    <origin rpy="1 0 0" xyz="0 1 0"/>
    <parent link="base_link"/>
    <child link="third_link"/>
    <mimic joint="base_to_second_joint" multiplier="1" offset="0"/>
  </joint>

</robot>"""

    params = {'robot_description': robot_desc}
    rsp = launch_ros.actions.Node(package='robot_state_publisher',
                                  executable='robot_state_publisher',
                                  output='both',
                                  parameters=[params])
    reset_description = ExecuteProcess(
        cmd=[
                [
                    FindExecutable(name="ros2"),
                    " param set ",
                    "/robot_state_publisher ",
                    "robot_description ",
                    "\"<robot name='test'><link name='base_link'/></robot>\"",

                ]
        ],
        shell=True,
    )

    return launch.LaunchDescription(
        [
            rsp,
            RegisterEventHandler(
                 OnProcessStart(
                     target_action=rsp,
                     on_start=[
                         LogInfo(msg="robot_state_publisher, resetting description"),
                         reset_description,
                     ],
                 )
             ),
        ]
    )
```
If you remove `<mimic joint="base_to_second_joint" multiplier="1" offset="0"/>` from the first description, it will not crash.

## Possible fix
See draft PR here: https://github.com/ros/robot_state_publisher/pull/212

doisyg:
Works for me to fix: https://github.com/ros/robot_state_publisher/issues/211
Not sure it is the cleanest approach through, opening as a draft to discuss first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,9 +95,20 @@ if(BUILD_TESTING)
 
   add_executable(test_two_links_change_fixed_joint test/test_two_links_change_fixed_joint.cpp)
   target_include_directories(test_two_links_change_fixed_joint PRIVATE ${GTEST_INCLUDE_DIRS})
-  target_link_libraries(test_two_links_change_fixed_joint PRIVATE ${GTEST_LIBRARIES} rclcpp::rclcpp tf2_ros::tf2_ros)
+  target_link_libraries(test_two_links_change_fixed_joint PRIVATE
+    ${GTEST_LIBRARIES}
+    ${rcl_interfaces_TARGETS}
+    rclcpp::rclcpp
+    tf2_ros::tf2_ros
+  )
   add_launch_test(test/two_links_change_fixed_joint-launch.py
     ARGS "test_exe:=$<TARGET_FILE:test_two_links_change_fixed_joint>")
+
+  add_executable(test_change_mimic_joint test/test_change_mimic_joint.cpp)
+  target_include_directories(test_change_mimic_joint PRIVATE ${GTEST_INCLUDE_DIRS})
+  target_link_libraries(test_change_mimic_joint PRIVATE ${GTEST_LIBRARIES} rclcpp::rclcpp ${rcl_interfaces_TARGETS})
+  add_launch_test(test/change_mimic_joint-launch.py
+    ARGS "test_exe:=$<TARGET_FILE:test_change_mimic_joint>")
 endif()
 
 ament_export_dependencies(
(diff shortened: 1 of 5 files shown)


clalancette:
Thanks for opening this.  For this PR, at the very least we should:

a) target it to `rolling` (we can backport from there), and
b) add in a test that shows that this fixes the problem

clalancette:
I've retargeted this to the `rolling` branch, and rebased.  I also changed this whole thing to a very explicit copy; while it is more verbose, it is clear what we are doing.  Finally, I added in a test here.  With all of that, this seems good to me so I'm going to mark it as "Ready for review" and run CI on it.  @doisyg any feedback from you is appreciated.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20033/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14497/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20761/)

doisyg:
Thanks for solving it Chris, and sorry for not following up (I was few days out).
This looks good and it works from me. Builds and runs fine under `iron`, our current production version (I would not mind a release of this fix to it at some point).
I am not sure I fully understand the root cause, but probably something like the shared pointer being deleted before the map is cleared. Anyway, it works now!

clalancette:
@Mergifyio backport iron humble

PR #215, opened 2024-01-12
Add joint states topic name parameter

mauricesvp:
Make the Joint States topic changeable.


clalancette:
We already have this functionality in a more generic way by using `remap` topics.  Please see the tutorial at http://docs.ros.org/en/rolling/Tutorials/Beginner-CLI-Tools/Introducing-Turtlesim/Introducing-Turtlesim.html#remapping for information on how to do it.

PR #220, opened 2024-06-26
Disable intra-process-comms for transient_local publisher

### LINKED ISSUE #219: Intra-process comm not supported due to `transient_local` durability
Currently, `robot_state_publisher` can be used as a composable node, but with Intra-Process disabled. It would be desiderable to allow its enablement.

Yadunund:
Fix #219

### DIFF
--- a/src/robot_state_publisher.cpp
+++ b/src/robot_state_publisher.cpp
@@ -129,10 +129,15 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
   tf_broadcaster_ = std::make_unique<tf2_ros::TransformBroadcaster>(this);
   static_tf_broadcaster_ = std::make_unique<tf2_ros::StaticTransformBroadcaster>(this);
 
+  auto publisher_options = rclcpp::PublisherOptions();
+  // Explicitly disable intra_process_comms for this publisher alone to allow
+  // for this node to run in a component container with intra_process_comms enabled.
+  publisher_options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
   description_pub_ = this->create_publisher<std_msgs::msg::String>(
     "robot_description",
     // Transient local is similar to latching in ROS 1.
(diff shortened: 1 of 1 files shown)


clalancette:
So I'm not certain of this, but I have the feeling that this should actually all work in `jazzy`/`rolling`.  This is, I believe we added `transient_local` support to intra-process comms before Jazzy.  Before we add in this workaround, can we make sure to try it there first?

roncapat:
Fair point @clalancette ... In my issue I haven't specified the distro I was using (from source, mostly Iron with some additional backports).

Will give it a try (maybe this patch could instead be targeted directly on 'iron' branch)?

src/robot_state_publisher.cpp
```diff
@@ -129,10 +129,15 @@ RobotStatePublisher::RobotStatePublisher(const rclcpp::NodeOptions & options)
   tf_broadcaster_ = std::make_unique<tf2_ros::TransformBroadcaster>(this);
   static_tf_broadcaster_ = std::make_unique<tf2_ros::StaticTransformBroadcaster>(this);
 
+  auto publisher_options = rclcpp::PublisherOptions();
+  // Explicitly disable intra_process_comms for this publisher alone to allow
+  // for this node to run in a component container with intra_process_comms enabled.
+  publisher_options.use_intra_process_comm = rclcpp::IntraProcessSetting::Disable;
```
sloretz, line 135: I'm surprised this is necessary.  Does this mean subscribers in the same process don't receive messages from publishers if the publishers have transient_local durability? I would have expected them to still get the messages through the rmw API.

Yadunund: In Iron and before an exception would be thrown by rclcpp. I wasn't aware that support for this was added in Jazzy/Rolling. So we can close this PR

clalancette:
I'm going to close this for now, as we don't need this on `rolling`.  If you'd like to see it on `iron` or `humble`, please feel free to open a PR to that branch.

PR #222, opened 2024-07-26
Added launch parameter for listener queue size.

JoeOnderko:
Problem:
I am publishing from a few nodes into the /joint_states topic and the subscriber in JointStateListener is not executing its callback function quick enough before the next JointState message is overwritten in the queue.

Fix:
I made the queue for the joint state listener configurable and set its default parameter to 1.


sloretz (approved):
Thanks for the PR!

